PR #4, opened 2022-06-01
Feat/delay: wrapper for delay()

DhruvaG2000:
If necessary, will also add basic instructions to README on how to build and test the included example.
Or we could maintain the entire documentation part on my blog and keep this repo code only?

Also, should we add doxygen based docs?


DhruvaG2000:
added instructions for building and running example here- https://dhruvag2000.github.io/Blog-GSoC22/docs/Hello-world.html

DhruvaG2000:
This PR requires code under arduino license hence this PR is to be DISCARDED. New PR will be made based on apache license that does not include any of Arduino Core API source files or header. They will be symlinked externally by the user at their local machine level

PR #6, opened 2022-06-29
Update README and license to Apache

DhruvaG2000:
Update project license from LGPL to Apache license

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,504 +1,201 @@
-                  GNU LESSER GENERAL PUBLIC LICENSE
-                       Version 2.1, February 1999
-
- Copyright (C) 1991, 1999 Free Software Foundation, Inc.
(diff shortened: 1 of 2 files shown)


beriberikix:
Let's also update the README at the same time. My proposed language:

Please note that the current license is Apache 2. Previously it was LGPL 2.1 but after careful review it was determined that no LGPL code or derivates was used and the more permissive license was chosen.

DhruvaG2000:
> Let's also update the README at the same time. My proposed language:
> 
> Please note that the current license is Apache 2. Previously it was LGPL 2.1 but after careful review it was determined that no LGPL code or derivates was used and the more permissive license was chosen.

Please check now if it looks ok

beriberikix (approved):

beriberikix (approved):

PR #8, opened 2022-07-01
feat/i2c: Add I2C wrapper functions

DhruvaG2000:
* Tested with a custom example for ADXL345 which can be found [here](https://github.com/DhruvaG2000/Arduino-Core-Zephyr/pull/5#issuecomment-1166949394)

* Follows latest licensing and directory structure

* uses ringBuffer

* TODO: Testing for multiple byte read/write

* TODO: Add i2c example if required

### DIFF
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -1,3 +1,9 @@
+/*
+ * Copyright (c) 2022 Dhruva Gole
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 #include "Wire.h"
 
 void arduino::ZephyrI2C::begin() {
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -1,0 +1,87 @@
+#include "Wire.h"
+
+void arduino::ZephyrI2C::begin() {
+  i2c_dev = DEVICE_DT_GET(DT_NODELABEL(i2c0));
+  ring_buf_init(&rxRingBuffer.rb, sizeof(rxRingBuffer.buffer), rxRingBuffer.buffer);
+}
+
+void arduino::ZephyrI2C::begin(uint8_t slaveAddr) {
+
+}
+
+void arduino::ZephyrI2C::end() {}
+
+void arduino::ZephyrI2C::setClock(uint32_t freq) {}
+
+void arduino::ZephyrI2C::beginTransmission(uint8_t address) { // TODO for ADS1115
+  _address = address;
+  usedTxBuffer = 0;
+}
+
+uint8_t arduino::ZephyrI2C::endTransmission(bool stopBit) {
+  int ret = i2c_write(i2c_dev, txBuffer, usedTxBuffer, _address);
+  if (ret) {
+    return 1; // fail
+  }
+  return 0;
+}
+
+uint8_t arduino::ZephyrI2C::endTransmission(void) { // TODO for ADS1115
+  return endTransmission(true);
+}
+
+size_t arduino::ZephyrI2C::requestFrom(uint8_t address, size_t len,
+                                       bool stopBit) {
+  int ret = i2c_read(i2c_dev, rxRingBuffer.buffer, len, address);
+  if (ret != 0)
+  {
+    printk("\n\nERR: i2c burst read fails\n\n\n");
+    return 0;
+  }
+  ret = ring_buf_put(&rxRingBuffer.rb, rxRingBuffer.buffer, len);
+  if (ret == 0)
+  {
+    printk("\n\nERR: buff put fails\n\n\n");
+    return 0;
+  }
+  return len;
+}
+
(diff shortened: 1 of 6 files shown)


DhruvaG2000:
For historical discussion on implementing i2c please refer: https://github.com/DhruvaG2000/Arduino-Core-Zephyr/pull/5

beriberikix (approved):
Defer to others for technical review

libraries/Wire/Wire.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+size_t arduino::ZephyrI2C::write(uint8_t data) {
+  txBuffer[usedTxBuffer++] = data;
+  return 1;
+}
+
+size_t arduino::ZephyrI2C::write(const uint8_t *buffer, size_t size) {
+  return size;
+}
```
alvarowolfx, line 67: We can implement this using memcpy to fill the txBuffer.
```
size_t arduino::ZephyrI2C::write(const uint8_t *buffer, size_t size) {  
  if (usedTxBuffer + size > 256) {
    size = 256 - usedTxBuffer;
  }
	memcpy(txBuffer + usedTxBuffer, buffer, size);
	usedTxBuffer += size;	
  return size;
}
```

libraries/Wire/Wire.cpp
```diff
@@ -0,0 +1,93 @@
...
+        if (ret == 0) {
+          printk("\n\nERR: buff empty\n\n\n");
+            return 0;
+        }
+		return (int)buf[0];
+  }
+  return rxBuffer[0];
+}
+
+int arduino::ZephyrI2C::available() { // TODO
```
alvarowolfx, line 82: We can implement the `available` function by checking if the ring buffer is not empty.
```
int arduino::ZephyrI2C::available() { 
  return !ring_buf_is_empty(&rxRingBuffer.rb);
}
```

libraries/Wire/Wire.cpp
```diff
@@ -0,0 +1,93 @@
...
+  return 1;
+}
+
+size_t arduino::ZephyrI2C::write(const uint8_t *buffer, size_t size) {
+  return size;
+}
+
+int arduino::ZephyrI2C::read() {
+  uint8_t buf[1];
+  if (ring_buf_peek(&rxRingBuffer.rb, buf, 1) > 0) {
```
alvarowolfx, line 71: We can change to use the internal `available` function

```
if (available() > 0) {
```

libraries/Wire/Wire.cpp
```diff
@@ -0,0 +1,93 @@
...
+        int ret = ring_buf_get(&rxRingBuffer.rb, buf, 1);
+        if (ret == 0) {
+          printk("\n\nERR: buff empty\n\n\n");
+            return 0;
+        }
+		return (int)buf[0];
+  }
+  return rxBuffer[0];
```
alvarowolfx, line 79: I think this should return `0` as we don't have data available in this case here.

libraries/Wire/Wire.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+int arduino::ZephyrI2C::available() { // TODO
+  return 1; 
+  }
+
+int arduino::ZephyrI2C::peek() { return 1; }   // TODO
```
alvarowolfx, line 86: `peek` can be implemented using the ring buffer peek function like you used before on the read function
```
int arduino::ZephyrI2C::peek() { 
  uint8_t buf[1];
  int bytes_read = ring_buf_peek(&rxRingBuffer.rb, buf, 1);
  if (bytes_read == 0){
    return 0;
  } 
  return (int)buf[0]; 
}
```

libraries/Wire/Wire.h
```diff
@@ -0,0 +1,63 @@
...
+  virtual int read();
+  virtual int peek();
+  virtual void flush();
+  virtual int available();
+  const struct device *i2c_dev;
+  static struct i2c_dt_spec bus;
+
+private:
+  int _address;
+  uint8_t txBuffer[256], rxBuffer[256];
```
alvarowolfx, line 50: We can remove this `rxBuffer` as we are using a Ring Buffer from zephyr.

alvarowolfx (changes requested):
I had some issues with `I2C` not being enabled, so I had to manually add `CONFIG_I2C` to my `prj.conf` file. Also the Ring buffer API is not enabled by default, so I had to add `CONFIG_RING_BUFFER`

One thing that we can do is to add to the lib `Kconfig`, so those things are enabled by default.

```
config ARDUINO_I2C
	bool "ARDUINO_API"
	select RING_BUFFER	
	select I2C
	default y
```

Another thing that I added here as I was testing, was an implementation for the Arduino `millis` function. 

```
unsigned long millis() { 
  return k_uptime_get();
}
```

alvarowolfx:
I tested this branch with the changes that I sent here using a BH1750 sensor and this library - https://github.com/claws/BH1750

The only change that I had to make to the BH1750 lib was to replace the `Serial.println` with `printk` statements as we don't support `Serial` yet.

Screen Shot 2022-08-01 at 16 13 36

PR #9, opened 2022-07-01
feat: add particle xenon pinmap

alvarowolfx:
I changed things so the pinmap file is added depending on the target board.

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -1,3 +1,9 @@
+/*
+ * Copyright (c) 2022 Dhruva Gole
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 #include "api/ArduinoAPI.h"
 
 /* Zephyr libraries */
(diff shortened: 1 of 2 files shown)


DhruvaG2000:
If there will be many more supported boards in the future, perhaps we should consider making a folder `cores/arduino/pinmaps/` that will include all the files like `particle_xenon_pinmap.h`. This will make the directory structure a bit more cleaner Imo.

beriberikix:
I recommend we follow the Arduino conventions and put different board in a [`variants` folder](https://github.com/arduino/ArduinoCore-mbed/tree/master/variants). It's not critical now but will be useful in the future when we want to integrate with the Arduino IDE, see the [platform specification](https://arduino.github.io/arduino-cli/0.24/platform-specification/#core-variants).

beriberikix:
@alvarowolfx when you get a chance can you adapt the implementation to the new structure?

DhruvaG2000:
> @alvarowolfx when you get a chance can you adapt the implementation to the new structure?

+1 
This would really help us from a new user perspective to understand difficulties that adding a new board may bring on table.

Issue #16, opened 2022-07-10
Move board definitions to `variants` folder


beriberikix:
Currently the pin mapping for the nano 33 ble is located in the Arduino core itself. Per the Arduino [platform specification](https://arduino.github.io/arduino-cli/0.24/platform-specification/#core-variants), these kind of files should be in their own directory under a `variants` folder (see [mbed](https://github.com/arduino/ArduinoCore-mbed/tree/master/variants) for an example.)

#9 should be updated once complete.

DhruvaG2000:
Is this the structure we are aiming for @beriberikix ?

```tree

└─❯ tree . -L 3
.
├── CMakeLists.txt
├── CODE_OF_CONDUCT.md
├── cores
│   ├── arduino
│   │   ├── api -> /home/dhruva/temp/ArduinoCore-API/api
│   │   ├── Arduino.h
│   │   ├── CMakeLists.txt
│   │   ├── main.cpp
│   │   └── zephyrCommon.cpp
│   └── CMakeLists.txt
├── Kconfig
├── LICENSE
├── README.md
├── samples
│   ├── blinky_arduino
│   │   ├── CMakeLists.txt
│   │   ├── prj.conf
│   │   ├── README.rst
│   │   ├── sample.yaml
│   │   └── src
│   ├── button_press_led
│   │   ├── CMakeLists.txt
│   │   ├── prj.conf
│   │   ├── README.rst
│   │   ├── sample.yaml
│   │   └── src
│   └── hello_arduino
│       ├── CMakeLists.txt
│       ├── prj.conf
│       ├── README.rst
│       ├── sample.yaml
│       └── src
├── variants
│   ├── ARDUINO_NANO33BLE
│   │   ├── arduino_nano_33_ble.overlay
│   │   └── arduino_nano_ble_sense_pinmap.h
│   └── CMakeLists.txt
└── zephyr
    └── module.yml
```

beriberikix:
Yes, that looks great!

beriberikix:
Closed with #17

PR #17, opened 2022-07-16
restruct: add variants folder

DhruvaG2000:
Solves https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/issues/16


DhruvaG2000:
[WIP] This code may not build yet, needs a few more additions.

beriberikix:
Sounds good!

Note: you can always create a draft PR before it is ready for review. https://github.blog/2019-02-14-introducing-draft-pull-requests/

DhruvaG2000:
Currently using a hard coded command to test on my system that may not work for everybody: 
```sh
west build -p -b arduino_nano_33_ble samples/basic/arduino-blinky -DZEPHYR_EXTRA_MODULES=/home/dhruva/zephyrproject/modules/lib/gsoc-2022-arduino-core -DDTC_OVERLAY_FILE=/home/dhruva/zephyrproject/modules/lib/gsoc-2022-arduino-core/variants/ARDUINO_NANO33BLE/arduino_nano_33_ble.overlay
```
The following changes are needed to eliminate the need to specify overlay path at build time command:

```diff
diff --git a/samples/blinky_arduino/CMakeLists.txt b/samples/blinky_arduino/CMakeLists.txt
index ba606d7..a75c14e 100644
--- a/samples/blinky_arduino/CMakeLists.txt
+++ b/samples/blinky_arduino/CMakeLists.txt
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0

 cmake_minimum_required(VERSION 3.20.0)
+set(DTC_OVERLAY_FILE $ENV{HOME}/zephyrproject/modules/lib/gsoc-2022-arduino-core/variants/ARDUINO_NANO33BLE/arduino_nano_33_ble.overlay)
 find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 project(blinky)
 ```
But with whatever name we spcify in the west.yml file:
```diff
diff --git a/west.yml b/west.yml
index f1dfa5a642..bffb71441a 100644
--- a/west.yml
+++ b/west.yml
@@ -25,6 +25,11 @@ manifest:
   #
   # Please add items below based on alphabetical order
   projects:
+      # Arduino API repository.
+    - name: Arduino-Core-Zephyr
+      path: modules/lib/Arduino-Zephyr-API
+      revision: feat/gpio
+      url: https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core
```
which would be ``Arduino-Zephyr-API`` in this case.

samples/button_press_led/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
 # SPDX-License-Identifier: Apache-2.0
 
 cmake_minimum_required(VERSION 3.20.0)
+
+set(DTC_OVERLAY_FILE $ENV{HOME}/zephyrproject/modules/lib/Arduino-Zephyr-API/variants/ARDUINO_NANO33BLE/arduino_nano_33_ble.overlay)
```
szczys, line 5: This ended up pointing to the wrong directory for me. I wonder if it would be okay to reference the Zephyr base and navigate in that way?

```
set(DTC_OVERLAY_FILE $ENV{ZEPHYR_BASE}/../modules/lib/Arduino-Zephyr-API/variants/ARDUINO_NANO33BLE/arduino_nano_33_ble.overlay)
```

szczys: Ah, the problem with this is that `ZEPHYR_BASE` isn't set unless you've run: 

```
source /your_path/zephyr/zephyr-env.sh
```

szczys: It's worth investigating to see if we can get the module path using `ZEPHYR_<MODULE_NAME>_MODULE_DIR` as described in this section of the docs: https://docs.zephyrproject.org/latest/develop/modules.html#module-name

samples/button_press_led/prj.conf
```diff
@@ -1,3 +1,4 @@
 CONFIG_GPIO=y
 CONFIG_CPLUSPLUS=y
-CONFIG_ARDUINO_API=y
\ No newline at end of file
+CONFIG_ARDUINO_API=y
+CONFIG_BOARD_ARDUINO_NANO_33_BLE_SENSE=y    # Ref. Arduino.h for supported boards
```
szczys, line 4: Following our conversation I tested this and found that the `-b` command at build time sets the board symbol for you

### After Building for Nano 33 BLE

```
$ grep -Ri arduino build/zephyr/.confi
CONFIG_BOARD="arduino_nano_33_ble"
CONFIG_BOARD_ARDUINO_NANO_33_BLE=y
# CONFIG_BOARD_ARDUINO_NANO_33_BLE_SENSE is not set
# CONFIG_BOOTLOADER_BOSSA_ARDUINO is not set
```

### After building for the Nano 33 BLE Sense

```
$ grep -Ri arduino build/zephyr/.config 
CONFIG_BOARD="arduino_nano_33_ble_sense"
# CONFIG_BOARD_ARDUINO_NANO_33_BLE is not set
CONFIG_BOARD_ARDUINO_NANO_33_BLE_SENSE=y
# CONFIG_BOOTLOADER_BOSSA_ARDUINO is not set

```

DhruvaG2000: I will definitely look into this and change the MACROs accordingly.. thanks for this insight!

cores/arduino/zephyrCommon.cpp
```diff
@@ -7,57 +7,28 @@
...
-    }
-  } else if (pinNumber < 100) {
-    if (pinMode == INPUT || pinMode == INPUT_PULLDOWN) {
-      gpio_pin_configure(DEVICE_DT_GET(DT_NODELABEL(gpio0)), pinNumber,
-                         GPIO_INPUT | GPIO_ACTIVE_LOW);
-    } else {
-      gpio_pin_configure(DEVICE_DT_GET(DT_NODELABEL(gpio0)), pinNumber,
-                         GPIO_OUTPUT);
-    }
+  if (pinMode == INPUT || pinMode == INPUT_PULLDOWN) { // input mode
```
szczys, line 10: If `INPUT_PULLDOWN` is selected, how are pulldowns activated? Do we assume that the pull-down flag has already been set in the overlay file, or is there some other mechanism?

samples/blinky_arduino/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
 # SPDX-License-Identifier: Apache-2.0
 
 cmake_minimum_required(VERSION 3.20.0)
+
+set(DTC_OVERLAY_FILE $ENV{HOME}/zephyrproject/modules/lib/Arduino-Zephyr-API/variants/ARDUINO_NANO33BLE/arduino_nano_33_ble.overlay)
```
szczys, line 5: I have not found a better way to do this, but it should be documented to help users who receive errors because they did not install in `/home/user/zerphyproject`

szczys (approved):

cores/arduino/zephyrCommon.cpp
DhruvaG2000: Not sure at this point what ``INPUT_PULLDOWN`` does in zephyr, I thought it was  a flag to make the gpio low state by default so it wouldn't have a floating value, but I think simple `INPUT` shud work as they have done here: https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/basic/button/src/main.c#L51

DhruvaG2000: > Do we assume that the pull-down flag has already been set in the overlay file, or is there some other mechanism?

I am not sure how GPIO_ACTIVE_LOW works underneath, if it does some DT Stuff or not...
But here's a useful line from the docs:

> Interrupts sensitive to pin logical level take into account GPIO_ACTIVE_LOW flag. If a pin was configured as Active Low, physical level low will be considered as logical level 1 (an active state), physical level high will be considered as logical level 0 (an inactive state).

My understanding was it would be better to set the pin `ACTIVE LOW` than leave it floating because I was not sure what or if zephyr has some default pull down in it.

samples/blinky_arduino/CMakeLists.txt
DhruvaG2000: will add documentation for this in README as well as my blog

cores/arduino/zephyrCommon.cpp
szczys: The `GPIO_ACTIVE_LOW` is an abstraction. Setting a pin to `HIGH` in your C code will turn that pin to the active level. Devicetree determines if that `HIGH` actually means voltage or ground based on `GPIO_ACTIVE_HIGH` or `GPIO_ACTIVE_LOW`.

There is a flag for pull-down resistors. So if the Arduino app wants to set a pull-down resistor, in Zephyr we should add the `GPIO_PULL_DOWN` flag when initializing the pin. https://docs.zephyrproject.org/latest/hardware/peripherals/gpio.html#c.GPIO_PULL_DOWN

szczys: Now I'm rethinking the topic of active level. I wonder if all GPIO pins should be set as active high? That way a `digitalWrite(pinNum, 1);` would drive the pin high. And for inputs, a V+ would be read as a 1 and ground as a 0. What are your thoughts?

DhruvaG2000: > I wonder if all GPIO pins should be set as active high

Yes, I think this would be the right thing to do

DhruvaG2000: Check https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/pull/17/commits/044b45912185c7559470bc0f6114fda9dfc6b51e

samples/blinky_arduino/CMakeLists.txt
DhruvaG2000: Done in README: https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/pull/17/commits/93deab35624d39d0688c202b8e348a06a018e912

beriberikix (approved):
Awesome! This will help us scale the number of boards we support. Lots of design choices under the covers based on our discussions and and ideas from Mike.

README.md
```diff
@@ -8,7 +8,9 @@ Arduino Core API module for zephyr leverages the power of Zephyr under an Arduin
 
 ## Adding Arduino Core API to Zephyr
 
-Add following entry to `west.yml` file in `manifest/projects` subtree of Zephyr:
+* **Pre requisites:** It is assumed that you have zephyrproject configured and installed on your system as per the official [Get Started Guide](https://docs.zephyrproject.org/latest/develop/getting_started/index.html). The recommended path to install is `~/zephyrproject` as specified in the guide. You may face path issues if you have zephyr installed in a custom path.
```
szczys, line 11:
```suggestion
* **Pre requisites:** It is assumed that you have zephyrproject configured and installed on your system as per the official [Get Started Guide](https://docs.zephyrproject.org/latest/develop/getting_started/index.html). The recommended path to install is `~/zephyrproject` as specified in the guide. If you have zephyr installed in a custom path you may need to make changes to the CMakeLists.txt file in the sample code directory when building these samples.
```

README.md
```diff
@@ -17,21 +19,20 @@ Add following entry to `west.yml` file in `manifest/projects` subtree of Zephyr:
...
 ```sh
 $ git clone git@github.com:arduino/ArduinoCore-API # Any location
 $ ln -s /<your>/<location>/ArduinoCore-API/api cores/arduino/.
 ```
 
-**Known Bugs:**
-While compiling with the ArduinoCore-API `WCharacter.h` produces many errors. It needs to be deleted from ArduinoAPI.h (it is unused in this port.) We are looking into resolving the issue.
-
+**Known Bug(s):**
+* While compiling with the ArduinoCore-API `WCharacter.h` produces many errors. It needs to be deleted from ArduinoAPI.h (it is unused in this port.) We are looking into resolving the issue.
```
szczys, line 35:
```suggestion
* While compiling with the ArduinoCore-API `WCharacter.h` produces many errors. The include of that file needs to be deleted from `cores/arduino/api/ArduinoAPI.h` (it is unused in this port.) We are looking into resolving the issue.

File: cores/arduino/zephyrCommon.cpp [hunk 7]
```diff
@@ -6,58 +6,29 @@
...
+    gpio_pin_configure_dt(arduino_pins[pinNumber],
+                          GPIO_INPUT | GPIO_PULL_DOWN | GPIO_ACTIVE_HIGH);
+  } else { // output mode
+    gpio_pin_configure_dt(arduino_pins[pinNumber],
+                          GPIO_OUTPUT_LOW | GPIO_PULL_DOWN | GPIO_ACTIVE_HIGH);
   }
 }
```
szczys, line 24: We're almost there. I suggest these changes: 

```suggestion
void pinMode(pin_size_t pinNumber, PinMode pinMode) {
  if (pinMode == INPUT) { // input mode
    gpio_pin_configure_dt(arduino_pins[pinNumber],
                          GPIO_INPUT | GPIO_ACTIVE_HIGH);
  } else if (pinMode == INPUT_PULLUP) { // input with internal pull-up
    gpio_pin_configure_dt(arduino_pins[pinNumber],
                          GPIO_INPUT | GPIO_PULL_UP | GPIO_ACTIVE_HIGH);
  } else if (pinMode == INPUT_PULLDOWN) { // input with internal pull-down
    gpio_pin_configure_dt(arduino_pins[pinNumber],
                          GPIO_INPUT | GPIO_PULL_DOWN | GPIO_ACTIVE_HIGH);
  } else if (pinMode == OUTPUT) { // output mode
    gpio_pin_configure_dt(arduino_pins[pinNumber],
                          GPIO_OUTPUT_LOW | GPIO_PULL_DOWN | GPIO_ACTIVE_HIGH);
  }
}
```
The reason for this is that the user should have the option of using their own external pull-up/pull-down resistor so we should not set the pull-down resistor by default, but only if the user code directs us to do so using the `INPUT_PULLDOWN` when setting pinmode.

Also, it is probably the same thing, but instead of defaulting to setting the pin as an output, we should test to make sure that's what the user asked for.

szczys (commented):
I made a code suggestion for pinMode changes, and a couple of tweaks to the README. This is looking great!

cores/arduino/zephyrCommon.cpp
szczys: Oops, looks like I missed the boat on this one. I'll open a PR with these changes. Thanks!

DhruvaG2000: Yes I agree with these suggestions.

README.md
DhruvaG2000: I am thinking we'll make a seperate PR to update the README's of all the samples.
Currently they are just ditto copy pasted from the zephyr samples so may not make much sense.
But we could also leave this as a `good-beginner-task` to perhaps attract potential contributors.

PR #25, opened 2022-08-07
Split Nano 33 BLE & Nano 33 BLE Sense

beriberikix:


DhruvaG2000:
I think we can then remove ``ARDUINO_NANO33BLE`` from variants as well ...
This will ofcourse require a path change across the CMakelists in all the samples as well.

DhruvaG2000:
Just tried building both nano ble and sense and both build fine! Without any changes in any cmakelists!

So I guess this PR LGTM

beriberikix:
Great! But let me make those changes anyway before merging 👍

DhruvaG2000 (approved):
Built and tested both variants, ``arduino_nano_33_ble_sense`` and ``arduino_nano_33_ble`` and since they pretty much have the same pin out, the examples worked just fine.
IMO this PR can be merged to dev.

DhruvaG2000:
> Great! But let me make those changes anyway before merging

yes, otherwise the samples may break.

PR #28, opened 2022-08-07
Nano ble cleanup

beriberikix:
Redoing this, includes a last typo fix


DhruvaG2000 (approved):
Built and tested locally, LGTM

PR #36, opened 2022-08-20
variants: add support for arduino_mkrzero

DhruvaG2000:
Adds support for the arduino mkr zero board. Tested with LED_Blink sample.
I am not sure about the inbuilt LED pin number though, hence this pinmap just uses D7 for
demo purposes

### DIFF
--- a/variants/arduino_mkrzero/arduino_mkrzero.overlay
+++ b/variants/arduino_mkrzero/arduino_mkrzero.overlay
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2022 Dhruva Gole <goledhruva@gmail.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+/ {
+	zephyr,user {
+		d0_gpios = <&arduino_mkr_header 0 0>;
+		d1_gpios = <&arduino_mkr_header 1 0>;
+		d2_gpios = <&arduino_mkr_header 2 0>;
+		d3_gpios = <&arduino_mkr_header 3 0>;
+		d4_gpios = <&arduino_mkr_header 4 0>;
+		d5_gpios = <&arduino_mkr_header 5 0>;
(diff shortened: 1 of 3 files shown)


DhruvaG2000:
@szczys can you help with inbuilt LED please?
I tried ``d22_gpios = < &portb  8 GPIO_ACTIVE_LOW>;`` but did not work

DhruvaG2000:
However when I add ``    <4  0 &portb 8 0>,  `` in the connector then the D4 pin works as internal LED... I am a little confused regarding what is hapenning.
Same when I add
```c
			   <22 0 &portb  8 0>;	/* BUILTIN LED */
```
It doesn't work when I do 
```c
led0 = <&arduino_mkr_header 22 0>;
```
in my overlay

Am I missing something here?

szczys:
@DhruvaG2000 

> I tried d22_gpios = < &portb 8 GPIO_ACTIVE_LOW>; but did not work

I am also confused. I looked at the schematic, the port and pin look correct but the flag should be `GPIO_ACTIVE_HIGH`.

image

However, when you look at the DTS file in Zephyr it's a very different story. They're using pin 17 for the LED:

https://github.com/zephyrproject-rtos/zephyr/blob/e7591615724894a2319ea2c09d7eb21b4263fbf0/boards/arm/arduino_zero/arduino_zero.dts#L32-L35

Maybe give that a try and see if it works? I suspect Arduino and Zephyr handle the pin mux differently which results in different naming scheme for the pins, but I'm only guessing.

DhruvaG2000:
I should point out here, that the link you sent is for the arduino_zero and
not the arduino_mkr_zero. There are changes in those 2 boards although I
guess the chip remains the same. I am in talks with @galak on the discord
channel, let's see if we can figure something out on this later perhaps,
for now we should probably not include either the inbuilt LED and advise
user to use an external one?

szczys (approved):
LGTM!

I don't have a board to test this on, but I was able to build samples.

PR #37, opened 2022-08-23
Add micros() and millis() functions

szczys:
Wraps Zephyr kernel timing calls in Arduino API. This completes the Arduino Time functions.

### DIFF
--- a/cores/arduino/zephyrCommon.cpp
+++ b/cores/arduino/zephyrCommon.cpp
@@ -38,3 +38,10 @@ PinStatus digitalRead(pin_size_t pinNumber) {
 void delay(unsigned long ms) { k_sleep(K_MSEC(ms)); }
 
 void delayMicroseconds(unsigned int us) { k_sleep(K_USEC(us)); }
+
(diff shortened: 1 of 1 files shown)


DhruvaG2000:
Could we perhaps add a sample to demonstrate this? I'll add it if the consensus is yes..
Otherwise this LGTM
Haven't got the chance to test it but it looks alright to me.

szczys:
These are standard Arduino functions. I tested them by adding this inside the loop of blinky_arduino:

```
printk("micros: %ld", micros());
printk("millis: %ld", millis());
```

My feeling is that they don't really belong in the blinky sample. We could do a "kitchen sink" sample that demonstrates everything. Or maybe some type of test app for that purpose?

DhruvaG2000:
Or else I feel we shouldn't over do on the samples either, I think these are pretty simple and standard in the arduino docs and any user can just google it and get to know how it is used

DhruvaG2000 (approved):
LGTM, Merging into dev

Issue #38, opened 2022-08-27
Serial.print fails


DhruvaG2000:
The wrapper for ``Serial.print(val >> i & 1, BIN);`` and ``Serial.print(i, HEX);`` is probably missing which causes build failures.

DhruvaG2000:
Refer: https://www.arduino.cc/reference/en/language/functions/communication/serial/print/

> Serial.print(78, BIN) gives "1001110"

Serial.print(78, OCT) gives "116"

Serial.print(78, DEC) gives "78"

Serial.print(78, HEX) gives "4E"

DhruvaG2000:
Adding more bugs:
```
[59/162] Building CXX object CMakeFiles/app.dir/src/main.cpp.obj
FAILED: CMakeFiles/app.dir/src/main.cpp.obj
[...]

PR #39, opened 2022-08-27
zephyrSerial: Add more print variants

DhruvaG2000:
This variant is capable of printing numbers and according to the base parameter.
    For eg. Serial.print(78, DEC) will print the decimal equivalent of 78 and
    Serial.print(78, HEX)  will print HEX and so on.

Solves https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/issues/38


cores/arduino/zephyrSerial.cpp
```diff
@@ -20,6 +24,32 @@ size_t arduino::ZephyrSerial::print(char ch){
     return print_char(ch, false);
 }
 
+size_t arduino::ZephyrSerial::print(const int val) {
+    printk("%d",val);
+    return sizeof(int);
+}
+
+size_t arduino::ZephyrSerial::print(double d) {
+    printk("%0.4f",d);
```
szczys, line 33:
```suggestion
    printk("%0.2f",d);
```

I think [according to the Arduino reference](https://www.arduino.cc/reference/en/language/functions/communication/serial/print/) the default floating point precision is 2 decimal places.

Also, we should have the same function but for `float` type parameters.

cores/arduino/zephyrSerial.cpp
```diff
@@ -20,6 +24,32 @@ size_t arduino::ZephyrSerial::print(char ch){
...
+    return sizeof(double);
+}
+
+size_t arduino::ZephyrSerial::print(const int val, const int base) {
+    if (base == 2) {   /* Todo: print Binary */
+        printk("%d", val);
+    }   else if (base == 16) {  /* print Hex value */
+        printk("%x",val);
+    }   else if (base == 8) {   /* Todo: print octal value */
+        printk("%x", val);
```
szczys, line 43: Is `%x` the formatter for octal? It would be `%o` for `printf`, but [I don't know if you can use that for `printk`](https://docs.zephyrproject.org/apidoc/latest/printk_8h.html#a768a7dff8592b69f327a08f96b00fa54).

cores/arduino/zephyrSerial.cpp
```diff
@@ -20,6 +24,32 @@ size_t arduino::ZephyrSerial::print(char ch){
...
+    }   else if (base == 10) {  /* print decimal value */
+        printk("%d", val);
+    }   else {
+        return EINVAL;
+    }
+
+    return 1;   /* Always return 1 byte */
```
szczys, line 50: This [should return](https://www.arduino.cc/reference/en/language/functions/communication/serial/print/#_returns) the number of bytes written. In this function, each conversion might use the following to do the actual printing:
```
char buffer[SOME_BUFFER_SIZE];
snprintk(buffer, sizeof(buffer), "%x", value);
printk("%s", buffer)
```

Then at the end of the function:

```
return strlen(buffer);
```

szczys (changes requested):

cores/arduino/zephyrSerial.cpp
szczys: You may consider implementing a `size_t arduino::ZephyrSerial::print(double d, const int precision )` precision function and calling that from this function with `2` as the precision. You can do an `snprintk` with the precision as a variable in order to create the format string passed to the `printk` statement:

```
snprintk(buffer, sizeof(buffer), "\%0.%df", precision);
printk(buffer, d);
```

szczys (approved):
Recent changes LGTM, thanks!

Issue #40, opened 2022-08-29
building external library fails with yield not implemented errors


DhruvaG2000:
Fix is to just:
```diff
diff --git a/cores/arduino/zephyrCommon.cpp b/cores/arduino/zephyrCommon.cpp
index a9c053b..7e47fcc 100644
--- a/cores/arduino/zephyrCommon.cpp
+++ b/cores/arduino/zephyrCommon.cpp
@@ -6,6 +6,9 @@

 #include "Arduino.h"

+void yield(void) {}
+//empty
+
 /*
  *  The ACTIVE_HIGH flag is set so that A low physical
  *  level on the pin will be interpreted as value 0.
```

PR #41, opened 2022-08-31
zephyrCommon: Implement yield()

DhruvaG2000:
* Fixes https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/issues/40
* yield(): causes the current thread to yield execution to another thread of the same or higher priority.
If there are no other ready threads of the same or higher priority, the routine returns immediately.
* Also replace include Arduino "" with <>.


szczys (approved):
LGTM

Issue #47, opened 2022-09-13
zephyr build warnings: zephyr.h is deprecated, include <zephyr/kernel.h> instead


DhruvaG2000:
Whenever I try to build a sample using latest zephyr, I get a bunch of the following errors:
```
/home/dhruva/zephyrproject/zephyr/include/zephyr/zephyr.h:13:2: warning: #warning "<zephyr/zephyr.h> is deprecated, include <zephyr/kernel.h> instead" [-Wcpp]
   13 | #warning "<zephyr/zephyr.h> is deprecated, include <zephyr/kernel.h> instead"
      |  ^~~~~~~
[51/164] Building CXX object zephyr/CMakeFiles/zephyr.dir/home/dhruva/zephyrproject/modules/lib/Arduino-Zephyr-API/cores/arduino/zephyrCommon.cpp.obj
In file included from /home/dhruva/zephyrproject/modules/lib/Arduino-Zephyr-API/cores/arduino/./Arduino.h:10,
                 from /home/dhruva/zephyrproject/modules/lib/Arduino-Zephyr-API/cores/arduino/zephyrCommon.cpp:7:
/home/dhruva/zephyrproject/zephyr/include/zephyr/zephyr.h:13:2: warning: #warning "<zephyr/zephyr.h> is deprecated, include <zephyr/kernel.h> instead" [-Wcpp]
   13 | #warning "<zephyr/zephyr.h> is deprecated, include <zephyr/kernel.h> instead"

```

PR #49, opened 2022-09-17
variants: bunch all includes in common file

DhruvaG2000:
Also fix https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/issues/47


szczys (approved):
The changes look good. I would suggest that you add to your commit message to indicate this matches the Zephyr naming convention change from zephyr.h to kernel.h

PR #51, opened 2022-09-19
zephyrCommon: Implement attachInterrupt()/detachInterrupt()

soburi:
This implementation depends on CONFIG_GPIO_GET_CONFIG
to keep the GPIO state already configured by pinMode().


soburi:
I'm testing with nrf52840 with this patch https://github.com/soburi/zephyr/tree/nrfx_gpio_get_config
(This patch needs more verification.)

DhruvaG2000 (changes requested):

samples/attach_interrupt/prj.conf
```diff
@@ -0,0 +1,4 @@
+CONFIG_GPIO=y
+CONFIG_CPLUSPLUS=y
```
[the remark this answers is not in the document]
soburi: Removed CONFIG_CPLUSPLUS and CONFIG_GPIO.

cores/arduino/zephyrCommon.cpp
```diff
@@ -6,6 +6,153 @@
...
+  struct gpio_port_callback* pcb = (struct gpio_port_callback*)cb;
+
+  for(uint32_t i=0; i<max_ngpios; i++) {
+    if (pins & BIT(i)) {
+      pcb->handlers[BIT(i)]();
+    }
+  }
+}
+
+void pinModeImpl(pin_size_t pinNumber, gpio_flags_t pinmode, voidFuncPtr callback, gpio_flags_t intmode)
```
DhruvaG2000, line 92: It seems like this PR is doing much more than just Implementing attachInterrupt()/detachInterrupt, and thus in my opinion deserves to have it's own commit that separates the modifications that have been made in the way that pinMode was earlier implemented.

DhruvaG2000 (changes requested):

cores/arduino/zephyrCommon.cpp
DhruvaG2000: Also if possible, the commit message should justify the use of new pinMode being better than earlier one and also the addition of ``GPIO_MODE_MASK`` instead of directly using each individual ``GPIO_XX`` flag as and when required?

soburi: As a result of redesign, I was able to eliminate this.
It's just attachInterrupt/detachInterrupt and their internal implementation.

soburi:
@DhruvaG2000 

Redesign based on https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/pull/60.

DhruvaG2000:
Any reason this is a draft PR? Please mark as ready if you want it to be reviewed,
Thanks.

PR #52, opened 2022-09-19
include kernel.h instead of zephyr.h

soburi:
The zephyr.h is already deprecated.
Use kernel.h instead of it.


soburi:
close, duplicated #49

PR #54, opened 2022-09-23
zephyrCommon: Implement analogRead()/analogReference() 

soburi:
The 'io-channels' node under the 'zephyr,user' defines
available ADC channels.

The 'io-channel-pins' defines the mapping of ADC channels
to pin numbers.  API looks up the 'io-channel-pins' when
querying the ADC channel by digital pin number.
So it should be in the same order as 'io-channels' node.

For original Arduino API compatibility, ADC resolution
should configure as 10-bit.

ADC that is on MCU supported by Zephyr exists
only 16-bit resolution currently.
So it supports resolutions up to 16-bit.

Signed-off-by: TOKITA Hiroshi <tokita.hiroshi@fujitsu.com>


DhruvaG2000 (commented):
This PR adds a big chunk of code to overlay files, hence for adding further variants it maybe useful to have documentation for what all is needed to get analog working in these overlays.
Do you think you can also add documentation for the same?

samples/analog_input/prj.conf
```diff
@@ -0,0 +1,4 @@
+CONFIG_GPIO=y
+CONFIG_ADC=y
```
DhruvaG2000, line 2: I think all this can be shifted into the project's global KConfig

DhruvaG2000 (changes requested):

samples/analog_input/README.rst
```diff
@@ -0,0 +1,47 @@
...
+
+For example, to build this sample for :ref:`arduino_nano_33_ble`:
+
+.. zephyr-app-commands::
+   :zephyr-app: samples/basic/arduino-threads
+   :board: arduino_nano_33_ble
+   :goals: build flash
+   :compact:
+
+Change ``arduino_nano_33_ble`` appropriately for other supported boards.
```
DhruvaG2000, line 47: This README needs updating.

DhruvaG2000 (changes requested):

samples/analog_input/prj.conf
soburi: Fixed it.

samples/analog_input/README.rst
soburi: I forgot it.😱

Update README did.

soburi:
@DhruvaG2000 
Rebase to newest main branch done.
Could you please review again?

soburi:
> Code looks good, but let me test this patch on real hardware before approving and merging.

Okay, No need to hurry

DhruvaG2000 (approved):
Built and tested on my arduino nano 33 ble sense, and it works OK.

PR #56, opened 2022-09-23
zephyrCommon: Implement analogWrite()

soburi:
The 'pwms' node under the 'zephyr,user' defines available PWM channels.

The 'pwm-pins' defines the mapping of PWM channels to pin numbers.
API looks up the 'pwm-pins' when querying the PWM channel by digital pin number.
So it should be in the same order as 'pwms' node.

PWM channels should configure and tied to the output pins.
The period property of the PWM channel is used as resolution as PWM.
For original Arduino API compatibility, it should be 255.
The period property and API argument determines the duty ratio.

---

I verified this PR with mkrzero and nano_33_ble.
Some pins of the mkrzero seem not to work correctly.
Only verified configuration only enableed.


samples/fade/src/app.cpp
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (c) 2022 TOKITA Hiroshi <tokita.hiroshi@fujitsu.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <Arduino.h>
+#include "zephyrSerial.h"
```
DhruvaG2000, line 8: including zephyrSerial shouldn't really be necessary as it is included in Arduino.h itself.

variants/nrf52840dk_nrf52840/nrf52840dk_nrf52840.overlay
```diff
@@ -23,5 +23,66 @@
...
+	pinctrl-1 = <&pwm0_sleep>;
+	pinctrl-names = "default", "sleep";
+};
+
+&pwm1 {
+	status = "okay";
+	pinctrl-0 = <&pwm1_default>;
+	pinctrl-1 = <&pwm1_sleep>;
+	pinctrl-names = "default", "sleep";
+};
```
DhruvaG2000, line 88: Having documentation as to what all needs to be added in overlays might help here as well.

DhruvaG2000 (changes requested):

samples/fade/src/app.cpp
soburi: Fixed it.

samples/fade/prj.conf
```diff
@@ -0,0 +1,4 @@
+CONFIG_CPLUSPLUS=y
+CONFIG_ARDUINO_API=y
+CONFIG_GPIO=y
```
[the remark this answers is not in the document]
soburi: Fixed it.

samples/fade/src/app.cpp
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#include <Arduino.h>
+
+const int led = 3;  // PWM output pin.
+const int increments = 5;
+const int wait_ms = 10;
+
+void setup() {
+  /* Pin that use as the PWM output must not configure by pinMode() */
```
DhruvaG2000, line 14:
```suggestion
  /* Pin that use as the PWM output need not be configured by pinMode() */
```

DhruvaG2000 (approved):
Built locally, and it does build.
Also flashed it on my nano 33 ble sense, and changed the sample a bit to rather use LED_BUILTIN.
LGTM.
Minor nitpicks, but otherwise can merge.

DhruvaG2000:
I thought I would have rebase access, but seems like I can't squash commits, please can you squash my last commit ?

soburi:
> I thought I would have rebase access, but seems like I can't squash commits, please can you squash my last commit in previous one? Also rebase on top of main please. Thank you

squash and update done.

DhruvaG2000 (approved):
LGTM!

PR #57, opened 2022-09-24
zephyrCommon: Add Print::write(const uint8_t*, size_t) default implementation

soburi:
When compile with enabling CONFIG_DEBUG option,
the Print::write virtual function failed to link.

This implementation is usually not used.
Because the override implementation that implements by the subclass will be used.

Signed-off-by: TOKITA Hiroshi <tokita.hiroshi@fujitsu.com>


DhruvaG2000 (changes requested):
In my opinion this should go into ``cores/arduino/zephyrSerial.cpp``.
At the moment we are not really very well integrating the Arduino Core API in serial aspect, so perhaps you may want to drop the ``arduino::Print`` and use ``arduino::ZephyrSerial::`` and also add the ``write`` in respective header file where we write the func prototypes.

soburi:
@DhruvaG2000 
> In my opinion this should go into `cores/arduino/zephyrSerial.cpp`.
The Print class is also used by (TCP) Client class, UDP class, and also user-defined classes via the Stream class.
I think it should not be placed in zephyrSerial because it has no relation to the Serial class.

I rewrited it to splitting into the new file named 'zephyrPrint.cpp'.
Please review it.

DhruvaG2000:
> I rewrited it to splitting into the new file named 'zephyrPrint.cpp'.

Pardon, I don't see the split, are you missing any other file?

cores/arduino/zephyrPrint.cpp
```diff
@@ -0,0 +1,20 @@
...
+ * It will be overridden by subclassese.
+ */
+size_t arduino::Print::write(const uint8_t *buffer, size_t size)
+{
+  size_t i;
+  for (i=0; i<size && write(buffer[i]); i++) {
+  }
+
+  return i;
+}
```
DhruvaG2000, line 20: Okay, so this seems to be taken from [here](https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/Print.cpp#L34-L42)
```C
size_t Print::write(const uint8_t *buffer, size_t size)
{
  size_t n = 0;
  while (size--) {
    if (write(*buffer++)) n++;
    else break;
  }
  return n;
}
```
but with for loop instead.

DhruvaG2000:
Seems like this file can in future have a similar structure to original [Print.cpp](https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/Print.cpp)
Perhaps we can even consider migrating ZephyrSerial entirely to use this file itself.
@szczys and @alvarowolfx  I would love to hear your thoughts on this

cores/arduino/zephyrPrint.cpp
soburi: One confirmation.
Is it okayto copy from GPL?

DhruvaG2000 (approved):
Looks like a good starting point to me.
Built locally, LGTM.

cores/arduino/zephyrPrint.cpp
DhruvaG2000: Not really, I think this is also okay.

DhruvaG2000:
Been too long since approved,  I do not see any nacks, hence merging.

PR #60, opened 2022-10-10
variants: Introduce a digital-pins array to define digital pin

soburi:
The `d0-gpios` node format makes it difficult to handle
digital pin configuration from DTS macros.

Introduce the simple array `digital-pins` instead of `g0-gpios` to define digital pin
to GPIO port and pin mapping.

PR #61, opened 2022-10-10
variants: Introduce the builtin-led-gpios node

soburi:
Introduce the `builtin-led-gpios` node to make possible to configuring
the builtin-led from dts.
This node uses to determine the value of the `LED_BUILTIN` macro.

Use the `led0` alias to determine the `LED_BUILTIN` if the `builtin-led-gpios`
is not defined.

The behavior can override by defining the `LED_BUILTIN` in pinmap.h.
It uses definitions that are defined in the header preferentially.

---

The PR is contineued from #60.
I will rebase if #60 merged.

### DIFF
--- a/documentation/variants.md
+++ b/documentation/variants.md
@@ -129,6 +129,34 @@ The following example instantiates `Wire` and `Wire2` with each `i2c0` and `i2c1
 };
 ```
 
+### Configure Builtin-LED
+
+The `builtin-led-gpios` node defines the Builtin-LED.
+This node defines the `LED_BUILTIN` value by looking up the `digital-pin-gpios`
+array to find the index of the pin.
+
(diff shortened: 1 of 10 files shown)


szczys (approved):
This is a nice improvement and the changes LGTM.

This PR should also include an update to the variant documentation on this topic: https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/blob/5c305c663a66cb8f74fc1f2d3a22eba0eb515694/documentation/variants.md?plain=1#L57-L84

soburi:
@szczys 

Thank you for reviewing.

> This PR should also include an update to the variant documentation on this topic:

I update the document.

DhruvaG2000 (approved):
This makes the overall code around builtin LEDs much more generic and future proof. Great addition!

Tested-by: Dhruva Gole

PR #64, opened 2022-10-15
zephyrSerial: Redesign to supporting hardware serial

soburi:
Add support for the hardware serial device. The implementation
using the UART interrupt, enabling CONFIG_UART_INTERRUPT_DRIVEN.

Instantiate as 'Serial' with UART device defined in 'serials' array
under zephyr,user node. If the 'serials' array is not defined,
try to instantiate with a UART device labeled 'arduino_serial'.
If even  'arduino_serial' does not define, it uses the stub
implementation that redirects to printk().

---
As a result of this PR, changes Serial output goes to UART1(exporse in Arduino header)  in nrf52840dk_nrf52840 case.


samples/serial_event/README.rst
```diff
@@ -0,0 +1,20 @@
...
+
+Building and Running
+********************
+
+Build and flash serial_event sample as follows,
+
+```sh
+$> west build -p -b arduino_nano_33_ble sample/serial_event/
+
+$> west flash --bossac=/home/$USER/.arduino15/packages/arduino/tools/bossac/1.9.1-arduino2/bossac
```
DhruvaG2000, line 19: close code ticks after this .. ```

DhruvaG2000 (changes requested):
Thanks for this PR,
I feel like you also need to add overlay for ``nrf52840dk_nrf52840``, rest you've covered.

soburi:
> I feel like you also need to add overlay changes for `nrf52840dk_nrf52840`, rest you've covered.

It is no needed.
Because nrf52840dk_nrf52840 defines `arduino_serial`, uses it.
Use exists definition if possible to use. (It is rule for reduce our work.)

samples/serial_event/README.rst
soburi: Fixed it.

DhruvaG2000 (approved):
Built locally, yet to test on HW, but code LGTM

DhruvaG2000:
This is a major PR in serial, and it would be awesome to have @szczys review it as well.

szczys (approved):
I read through the commit and this LGTM. I didn't have the chance to do a deep dive into everything, @DhruvaG2000 do you have any remaining concerns with this PR?

DhruvaG2000:
> do you have any remaining concerns with this PR?

None, Thanks alot @soburi for this PR and Mike for reviewing! Apologies for the delay in merging.

PR #66, opened 2022-11-20
GPIO cells use to PWM and ADC pin assigning

soburi:
Change to use GPIO cells to assign PWM and ADC pins.
This change makes it easier to manage because unifying
the notation with `digital-pin-gpios`.
Deleting `pwm-pins` and `io-channel-pins` nodes, and create
`pwm-pin-gpios` and `adc-pins-gpios`.


soburi:
@DhruvaG2000 

This is last patch in series of overlay improvements.
Since it includes specification changes, I would appreciate it if you could release it again after merging.

DhruvaG2000 (approved):
LGTM

PR #67, opened 2022-11-20
CI/CD: setup docker image build check

DhruvaG2000:
Creates a CI CD workflow to build the next branch for every push  / pull request.
This will ensure that the next branch always builds and is in a build-able state. Thus just before merging into main we will have public record of the module being built.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,41 @@
+name: Build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: zephyrprojectrtos/ci:latest
+    env:
+      CMAKE_PREFIX_PATH: /opt/toolchains
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          path: Arduino-Zephyr-API
+
+      - name: Initialize
+        working-directory: Arduino-Zephyr-API
+        run: |
+          west init -m https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core.git
+          west update
+          git clone https://github.com/arduino/ArduinoCore-API.git ArduinoCore-API
+          sed '/WCharacter.h/ s/./\/\/ &/' ArduinoCore-API/api/ArduinoAPI.h > ArduinoCore-API/api/tmpArduinoAPI.h
+          mv ArduinoCore-API/api/tmpArduinoAPI.h ArduinoCore-API/api/ArduinoAPI.h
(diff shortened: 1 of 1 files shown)


beriberikix:
This is a pretty large PR and looks like it's pulling in ALL of Zephyr's CI checks. I don't think we really need as this is just a module.

What if we start small with a simple build check for one target device?

DhruvaG2000:
> This is a pretty large PR and looks like it's pulling in ALL of Zephyr's CI checks. I don't think we really need as this is just a module.
> 
> What if we start small with a simple build check for one target device?

yes I agree, I will remove unnecessary scripts and rebase / create a new PR.

beriberikix:
I've been maintaining [size-optimized containers](https://github.com/beriberikix/zephyr-docker) for local development and CI which could be a good basis for our use case. I haven't documented how to use them in CI but it should be straightforward.

DhruvaG2000:
Thanks for that reference @beriberikix , I will check it out.
Meanwhile the example-application given by Zephyr seemed adequate and easy to me so I decided to give it a shot,
But as you can see it fails:

```sh
<command-line>: fatal error: /__w/gsoc-[20](https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/actions/runs/3508707301/jobs/5877301749#step:5:21)22-arduino-core/gsoc-2022-arduino-core/zephyr/../modules/lib/Arduino-Zephyr-API/variants/arduino_nano_33_ble/arduino_nano_33_ble.overlay: No such file or directory
```

This is because we have the name "Arduino-Zephyr-API" hardcoded in our CMakelists.

Good thing is atleast there is progress on the fact that the existing zephyr docker container seems to work well and atleast reach the west build command, and ALSO find the project sample.

However we need to figure a way to somehow do 2 things,
1. We need to install the actual Arduino Core API somehow (our install.sh script) 
2. Also need to find a way to make the docker container have this project inside the proper path ie.``modules/lib/Arduino-Zephyr-API/``

DhruvaG2000:
@beriberikix  Kindly review this, there is nothing to really test here other than the CI Build passes and you can see the logs if you click on the "Details" for Build/ Build under checks.
Just for reference one of the passing logs can be seen here:  https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/actions/runs/3609412749/jobs/6082591758

Failing logs just for demo purposes till this PR merges, can be found here:
https://github.com/DhruvaG2000/arduino-core-testing/actions/runs/3609437467/jobs/6082631266
Where I have deliberately added an error in the code.

beriberikix (approved):

PR #71, opened 2022-12-18
samples: hello arduino: Clean up

DhruvaG2000:
Cleanup hello_arduino sample by removing comments that no longer apply and also removing unnecessary amount of prints.

Closes https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/issues/38

### DIFF
--- a/samples/hello_arduino/src/app.cpp
+++ b/samples/hello_arduino/src/app.cpp
@@ -3,27 +3,13 @@
  */
 
 #include <Arduino.h>
-#include "zephyrSerial.h"
-
 
 void setup() {
-  // put your setup code here, to run once:
-  Serial.begin(115200); // dummy as of now, need to study and refer https://docs.zephyrproject.org/latest/hardware/peripherals/uart.html
+	// put your setup code here, to run once:
+	Serial.begin(115200);
 }
 void loop() {
-  char c = 'D';
-  size_t ret1;
-  size_t ret2;
-  ret1 = Serial.print(c);
-  ret2 = Serial.println("Hello, World!");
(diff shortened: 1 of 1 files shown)


samples/hello_arduino/src/app.cpp
```diff
@@ -3,27 +3,13 @@
...
-  ret1 = Serial.print("My letter is: ");
-  ret2 = Serial.println(c);
-  printk("Sizes: %d %d\n", ret1, ret2);
-  Serial.println();
-  char myString[] = "Will it print?";
-  ret1 = Serial.println(myString);
-  printk("Size: %d \n\n\n", ret1);
-  delay(1000); // 1 second delay
+	// loop function for continuously executing code
+	Serial.println("Hello World!");
```
soburi, line 13: It is good as "hello world" example.

The original was also work as formatting test.
Is not need to consider?
(such like https://www.arduino.cc/reference/en/language/functions/communication/serial/print/)

soburi (commented):
It is good as "hello world" example.
The original was also work as formatting test.
Is not need to consider?
(Such like as https://www.arduino.cc/reference/en/language/functions/communication/serial/print/.)

DhruvaG2000:
> It is good as "hello world" example. The original was also work as formatting test. Is not need to consider? (Such like as https://www.arduino.cc/reference/en/language/functions/communication/serial/print/.)

I agree that we should have a way to demo other Serial.print variants as well.
Let me then rather than delete all other lines make them look neater.

DhruvaG2000:
Closing this- because this sample actually demonstrates a lot of useful stuff that our API can do so let's keep as is.

PR #74, opened 2023-08-04
Changes necessary for building samples on Windows.

ddeletic:
A minor modification to allow samples to build on Windows.


soburi:
Zephyr not define `DTC_OVERLAY_FILE` in build system.
So, we have no way to select overlay file to load in build system at this time.
(This PR may became problem when this repository integrate to main line.)
It must pass with `-DDTC_OVERLAY_FILE` externally as option.
For example, `west build ... -- -DDTC_OVERLAY_FILE=[overlay file path]`.
Is it not worked correctly?

DhruvaG2000:
> Zephyr not define DTC_OVERLAY_FILE in build system.

@soburi , yes this was indeed one of the challenges I faced in the making of this project. You can refer to my blog [here](https://dhruvag2000.github.io/Blog-GSoC22/#hurdles) to read more on how I solved this issue.

But yes @ddeletic  is right here, we can make use of `DTC_OVERLAY_FILE` in build time

soburi:
@ddeletic @DhruvaG2000 
Currently, this module is not integrated into the mainline, so I think that's fine for now. However, if it is integrated into the mainline, it will be necessary to reconsider at some point.

soburi (approved):

.gitignore
```diff
@@ -0,0 +1,4 @@
+samples/blinky_arduino/build/
+samples/hello_arduino/build/
+samples/fade/build/
+samples/button_press_led/build/
```
DhruvaG2000, line 4: do we really need this? On linux atleast I don't see any build folder appear in any `samples/`

DhruvaG2000:
@soburi , please excuse me but I don't still understand your exact concern?
Is the way it is currently implemented non ideal? Kindly help me understand the potential issues it could cause

DhruvaG2000 (changes requested):
I agree with the general idea of this PR, however please follow the commit guidelines from zephyr:

https://docs.zephyrproject.org/latest/contribute/guidelines.html#developer-certification-of-origin-dco

To summarise:
1. Please use an [imperative mood in the commit message](https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/)
2. Add some content to the body, empty commit body doesn't look good.
3. Add your signoff as per the Zephyr DCO requirements I mentioned earlier.

Thanks for contributing!

soburi:
@DhruvaG2000 

Simply because it except to this variable set from the command line.
If the command line gives the DTC_OVERLAY_FILE option, it will not add but overwrite the value set in CMakeLists.txt. It may cause incorrect behavior.

At the moment, Only some files that are processed in the `cmake/modules/configuration_files.cmake` can automatically load the overlay. 

I think we need to rethink cmake extensions when integrating.
(I think the way of the handling shield will be helpful.)

DhruvaG2000:
Ah yes got it, when someone wants to build using west build passing another
overlay other than the arduino dt overlay.. it will cause issues likely.
Maybe something like a += or something may help?
Let's think about it.

.gitignore
DhruvaG2000: No it's not about my `gitignore` , I just don't see the `build` folder appear in any `samples/`. Can you tell me what command you use to build samples that it is outputting to a build folder inside `samples/xyz/build`?
I see build only come up in the root location of this repo. We can add `build` to `.gitignore` if required.

DhruvaG2000: Or maybe it's a windows thing...?

soburi:
@ddeletic 

This way violates Zephyr's expected use of DTC_OVERLAY_FILE.
So we should not show a bad example in examples.
However, this repository has not yet merged, so we don't need to fix it now.
In this case, documentation is a better way to help beginners.

If you want to solve it implicitly, I think proposing a comprehensive solution to handle overlay files is better.
(I think it will probably be a set of specs and a framework written in cmake. It is a bit hard work.)

.gitignore
DhruvaG2000: OK, I suggest keeping these changes in a seperate commit, as they are unrelated to the "fix windows..." commit.

DhruvaG2000 (changes requested):

DhruvaG2000 (approved):

Issue #77, opened 2023-08-19
nrf52840dk build failure


DhruvaG2000:
The nrf52840dk build fails for all samples with following errors:

```
-- The C compiler identification is GNU 12.2.0
-- The CXX compiler identification is GNU 12.2.0
-- The ASM compiler identification is GNU
-- Found assembler: /home/dhruva/zephyr-sdk-0.16.1/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc
-- Configuring done (1.9s)
-- Generating done (0.0s)
-- Build files have been written to: /home/dhruva/zephyrproject/modules/lib/Arduino-Zephyr-API/build
-- west build: building application
[1/179] Preparing syscall dependency handling

[2/179] Generating include/generated/version.h
-- Zephyr version: 3.4.99 (/home/dhruva/zephyrproject/zephyr), build: v3.4.0-1121-g76663444c1fa
[53/179] Building CXX object zephyr/CMakeFiles/zephyr.dir/home/dhruva/zephyrproject/modules/lib/Arduino-Zephyr-API/cores/arduino/zephyrSerial.cpp.obj
FAILED: zephyr/CMakeFiles/zephyr.dir/home/dhruva/zephyrproject/modules/lib/Arduino-Zephyr-API/cores/arduino/zephyrSerial.cpp.obj 
[...]

PR #82, opened 2023-08-29
zephyrSerial: Fix misdetection arduino_serial1

DhruvaG2000:
This is a bug fix, that fixes build failure for nRF52940DK https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/issues/77

DT_NODE_EXISTS value is true even if the status is disabled, so it caused misdetection.
Use DT_NODE_HAS_STATUS to detect the status of
the arduino_serial1 node is ok.


DhruvaG2000:
@beriberikix , @szczys  @alvarowolfx  ,
I am merging this as it's been pending for 2 weeks now, and no objections have been raised.

PR #83, opened 2023-09-16
workflow: Fix the PR review CI

DhruvaG2000:
Improving checkpatch to function properly, as it earlier did not catch errors properly. One such incident was brought to light when a missing SOB went undetected.
Ensure that such errors are caught when the PRs are raised itself.


DhruvaG2000:
Example of a bad commit message being flagged here:
https://github.com/DhruvaG2000/arduino-core-testing/pull/9

DhruvaG2000:
Merging as no NAK's received.

PR #86, opened 2023-10-17
variants: remove intermediate variants.h

soburi:
* Move common part of variants/variants.h to core arduino/Arduino.h and remove variants/variants.h.
* Rename ..._pinmap.h to variants.h.
* Remove no needed includes.


variants/nrf52840dk_nrf52840/variants.h
```diff
@@ -0,0 +1,5 @@
+/*
+ * Copyright (c) 2022 Mike Szczys
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
DhruvaG2000, line 5: In cases where we don't do anything in variants.h do we just want to remove them altogether then?

soburi: I tried it, but the empty file was rejected by the CI check. 😨

cores/arduino/Arduino.h
```diff
@@ -4,12 +4,101 @@
...
+
+#ifdef CONFIG_ADC
+
+#define AN_ENUMS(n, p, i) A ## i = DIGITAL_PIN_GPIOS_FIND_PIN( \
+		DT_REG_ADDR(DT_PHANDLE_BY_IDX(DT_PATH(zephyr_user), p, i)),        \
+		DT_PHA_BY_IDX(DT_PATH(zephyr_user), p, i, pin)),
+enum analogPins { DT_FOREACH_PROP_ELEM(DT_PATH(zephyr_user),
+				       adc_pin_gpios, AN_ENUMS) };
+
+#endif
```
DhruvaG2000, line 101: Why are you making your changes in the middle here? Some of the #includes have moved to the bottom, making it hard to spot them.

variants/nrf52840dk_nrf52840/variants.h
DhruvaG2000: In that case I am thinking, should we get rid of variants.h altogether?
I don't think they serve a very important purpose looking at them so far, just a few #define for LED's is something we can do away with is what I feel.

cores/arduino/Arduino.h
soburi: The original code is in head of variants.h.
So I put the code before including variants.h to keep original order.

DhruvaG2000:
BTW, can you update the branch with latest next if you're pushing any new changes?

variants/nrf52840dk_nrf52840/variants.h
soburi: I think there still exists the possibility to implement platform-specific definitions.
It may bit too conservative, but it is better to keep it, in my opinion.

DhruvaG2000 (approved):

PR #92, opened 2024-05-11
Add support for beagleconnect_freedom

Ayush1325:
- Test blinky_arduino
- Test hello_arduino

Not sure how I should number the mikrobus digital pins (or if I even should).

### DIFF
--- a/variants/beagleconnect_freedom/beagleconnect_freedom.overlay
+++ b/variants/beagleconnect_freedom/beagleconnect_freedom.overlay
@@ -1,0 +1,14 @@
+/*
+ * Copyright (c) 2024 Ayush Singh <ayush@beagleboard.org>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+ / {
+	zephyr,user {
+		digital-pin-gpios = <&gpio0 18 GPIO_ACTIVE_HIGH>;	/* 2.4GHz TX/RX */
+		builtin-led-gpios = <&gpio0 18 GPIO_ACTIVE_HIGH>;	/*  2.4GHz TX/RX  */
+		serials = <&uart0 &uart1>;
+		i2cs = <&i2c0>;
+	};
+ };
--- a/variants/beagleconnect_freedom/variants.h
+++ b/variants/beagleconnect_freedom/variants.h
@@ -1,0 +1,9 @@
+/*
+ * Copyright (c) 2024 Ayush Singh <ayush@beagleboard.org>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#define LED_BUILTIN 0
+
--- a/documentation/variants.md
+++ b/documentation/variants.md
@@ -9,6 +9,7 @@ Once completed, continue here by adding a variant for your custom board.
 - [X] Arduino Nano ble sense 33
 - [X] Arduino Nano ble 33
 - [X] Arduino Nano 33 iot
+- [X] Beagleconnect Freedom
 
 ## Planned Support: (TODO)
 - [ ] Particle Xenon


variants/beagleconnect_freedom/variants.h
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2024 Ayush Singh <ayush@beagleboard.org>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* All the pins that are 100 + x are gpio1 pins and < 100 are in gpio0 */
```
DhruvaG2000, line 7: This comment doesn't make sense anymore and needs to be cleaned up.
It seems to be a remnant of commit e8e3c6294 ("restruct: add variants folder")

variants/beagleconnect_freedom/beagleconnect_freedom.overlay
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (c) 2024 Ayush Singh <ayush@beagleboard.org>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+ / {
+	zephyr,user {
+		digital-pin-gpios = <&gpio0 18 GPIO_ACTIVE_HIGH>;	/* 2.4GHz TX/RX */
+		builtin-led-gpios = <&gpio0 18 GPIO_ACTIVE_HIGH>;	/*  2.4GHz TX/RX  */
```
DhruvaG2000, line 10: You  can also add a ``adc-pin-gpios =    `` right?

variants/beagleconnect_freedom/beagleconnect_freedom.overlay
```diff
@@ -0,0 +1,14 @@
...
+		serials = <&uart0 &uart1>;
+		i2cs = <&i2c0>;
```
DhruvaG2000, line 12: Also missing ``pwms = `` ?

DhruvaG2000 (changes requested):

variants/beagleconnect_freedom/beagleconnect_freedom.overlay
Ayush1325: `adc-pin-gpios` is supposed to be defined for analog input pins. The pins in bcf are for mikrobus connector. So it would only make sense to add `adc-pin-gpios` if we are adding `digital-pin-gpios` as well. The reason for omission is that I am not sure how to name them since they are not really arduino header (or even numbered).

variants/beagleconnect_freedom/beagleconnect_freedom.overlay
Ayush1325: I am not sure if anything is supposed to go into pwms for bcf? I modeled the overlay after `cc3220sf_launchxl` which also does not contain `pwms`. The device tree for bcf also does not seem to have any pwm nodes.

DhruvaG2000: I was just looking at https://www.mikroe.com/img/cms/mikrobus-image-1.jpg hence wondered if we need pwm, but if zephyr itself doesn't support pwm for beagleconnect today then ok to skip.

DhruvaG2000 (changes requested):
One small request, can you also please update https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/blob/next/documentation/variants.md ?

Ayush1325:
> One small request, can you also please update https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/blob/next/documentation/variants.md ?

Hi. I have added beagleconnect freedom to supported list. However, it seems that the supported list is outdated and some of the planned boards are also already supported.
Would you like me to edit those in this PR as well or is something missing. Alternatively I can create a new PR for that (or you can add them yourself later).

DhruvaG2000 (approved):
LGTM, Will update the supported variants list later on... This much is good for now.

variants/beagleconnect_freedom/beagleconnect_freedom.overlay
jadonk: Why not all the other digital pins? 

https://openbeagle.org/beagleconnect/zephyr/arduino-core/-/commit/1ce0b0bfca16ffc9eb781f263525347e0b42cef5

PR #94, opened 2024-05-14
beagleconnect_freedom: Add digital, analog, pwm pins

Ayush1325:
- Add pin layout described here: https://openbeagle.org/beagleconnect/zephyr/arduino-core/-/commit/1ce0b0bfca16ffc9eb781f263525347e0b42cef5
- Also remove LED_BUILTIN. Seems the dts entry is sufficient.

### DIFF
--- a/variants/beagleconnect_freedom/beagleconnect_freedom.overlay
+++ b/variants/beagleconnect_freedom/beagleconnect_freedom.overlay
@@ -6,7 +6,44 @@
 
  / {
 	zephyr,user {
-		digital-pin-gpios = <&gpio0 18 GPIO_ACTIVE_HIGH>;	/* 2.4GHz TX/RX */
+		digital-pin-gpios = 
+				    <&gpio0 16 GPIO_ACTIVE_HIGH>, /* D0  - MB1 INT */
+				    <&gpio0 20 GPIO_ACTIVE_HIGH>, /* D1  - MB2 INT */
+				    <&gpio0 17 GPIO_ACTIVE_HIGH>, /* D2  - MB1 PWM */
+				    <&gpio0 28 GPIO_ACTIVE_HIGH>, /* D3  - MB1 CS - A5 */
+				    <&gpio0 21 GPIO_ACTIVE_HIGH>, /* D4  - MB2 UART1 RX */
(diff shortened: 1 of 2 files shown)


Ayush1325:
Cc: @jadonk

DhruvaG2000 (changes requested):
@Ayush1325 The commit message feels inadequate to me:

> - Add pin layout described here:
https://openbeagle.org/beagleconnect/zephyr/arduino-core/-/commit/1ce0b0bfca16ffc9eb781f263525347e0b42cef5


You need to add reasoning as to why it was chosen the way it is. Is there any pinout or diagram somewhere that can justify this mapping?

DhruvaG2000:
My main concern here is a person comin new to the platform should be able to quickly make sense of what pin mapping you've created

jadonk:
Thanks. Looks good to me. Do you have a test procedure?

Ayush1325:
> Thanks. Looks good to me. Do you have a test procedure?

Can you provide some clarification regarding how the pin numbering was chosen. Analog numbering seems simple but not sure about digital pin order.

jadonk:
> > Thanks. Looks good to me. Do you have a test procedure?
> 
> Can you provide some clarification regarding how the pin numbering was chosen. Analog numbering seems simple but not sure about digital pin order.

Sure.

I looked around for some examples of Arduino Click Shields such that Arduino examples that used mikroBUS boards might "just work", but I found that there wasn't much consistency between the Shields. I chose what I thought was both reasonably popular, affordable and included a reasonably good set of pins and documentation.

I want with [MikroE Arduino MKR Click Shield](https://www.mikroe.com/arduino-mkr-click-shield)

image

image

I tried to match up the equivalent Arduino pin with the pins on the 2 mikroBUS ports on BeagleConnect Freedom. My thinking was that if it worked on an Arduino MKR with the MikroE Arudino MKR Click Shield, then it should work with BeagleConnect Freedom. That should provide a great test resource.

Can you add this information to the documentation in a reasonable way?

Ayush1325:
@jadonk Thanks for the input. I do not think there is any board specific docs currently, but we can certainly add a section in our own docs regarding this. 

@DhruvaG2000 I think this provides the context for the pin layout?

DhruvaG2000 (approved):
Yep, looks good now.

PR #98, opened 2024-06-05
beagleconnect_freedom: Add pwm support

Ayush1325:
- Replace pwm_set_cycles with pwm_set_pulse_dt.
- Add check to ensure that pwm pin is ready
- Fix intendation
- Fix incorrect PWM MB 2 pin number
- Add pwms property

### DIFF
--- a/cores/arduino/zephyrCommon.cpp
+++ b/cores/arduino/zephyrCommon.cpp
@@ -217,6 +217,10 @@ void analogWrite(pin_size_t pinNumber, int value)
 {
   size_t idx = pwm_pin_index(pinNumber);
 
+  if (!pwm_is_ready_dt(&arduino_pwm[idx])) {
+    return;
+  }
+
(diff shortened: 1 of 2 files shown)


cores/arduino/zephyrCommon.cpp
```diff
@@ -217,6 +217,10 @@ void analogWrite(pin_size_t pinNumber, int value)
 {
   size_t idx = pwm_pin_index(pinNumber);
 
+	if (!pwm_is_ready_dt(&arduino_pwm[idx])) {
```
soburi: Please fix indentation

soburi (approved):

DhruvaG2000 (approved):
LGTM, let me know if this is tested and good to merge.

Ayush1325:
@DhruvaG2000 So, the thing is that this works as well as pwm works for bcf in zephyr (which means not very well). However, the problem seems to be in the zephyr driver (and not on arduino side) and thus, it should be fine to merge this.

PR #100, opened 2024-06-06
zephyrCommon: Add tone and noTone 

Ayush1325:
- Add implementation of tone and noTone
- Add sample. Taken from https://github.com/arduino/arduino-examples/blob/main/examples/02.Digital/toneMelody/toneMelody.ino
- Tested on bcf

### DIFF
--- a/cores/arduino/zephyrCommon.cpp
+++ b/cores/arduino/zephyrCommon.cpp
@@ -210,6 +210,51 @@ PinStatus digitalRead(pin_size_t pinNumber) {
   return (gpio_pin_get_dt(&arduino_pins[pinNumber]) == 1) ? HIGH : LOW;
 }
 
+struct k_timer arduino_pin_timers[ARRAY_SIZE(arduino_pins)];
+struct k_timer arduino_pin_timers_timeout[ARRAY_SIZE(arduino_pins)];
+
+void tone_expiry_cb(struct k_timer *timer) {
+  const struct gpio_dt_spec *spec = (gpio_dt_spec*)k_timer_user_data_get(timer);
+  gpio_pin_toggle_dt(spec);
+}
+
+void tone_timeout_cb(struct k_timer *timer) {
+  pin_size_t pinNumber = (pin_size_t)(uintptr_t)k_timer_user_data_get(timer);
+  noTone(pinNumber);
+}
+
+void tone(pin_size_t pinNumber, unsigned int frequency, unsigned long duration) {
+  struct k_timer *timer = &arduino_pin_timers[pinNumber];
+  const struct gpio_dt_spec *spec = &arduino_pins[pinNumber];
+  k_timeout_t timeout;
+
+  pinMode(pinNumber, OUTPUT);
+
+  if (frequency == 0) {
+    gpio_pin_set_dt(spec, 0);
+    return;
+  }
+
+  timeout = K_NSEC(NSEC_PER_SEC / (2 * frequency));
+
+  k_timer_init(timer, tone_expiry_cb, NULL);
+  k_timer_user_data_set(timer, (void*)spec);
+  gpio_pin_set_dt(spec, 1);
+  k_timer_start(timer, timeout, timeout);
+
+  if(duration > 0) {
+    timer = &arduino_pin_timers_timeout[pinNumber];
+    k_timer_init(timer, tone_timeout_cb, NULL);
+    k_timer_user_data_set(timer, (void*)(uintptr_t)pinNumber);
+    k_timer_start(timer, K_MSEC(duration), K_NO_WAIT);
+  }
+}
+
+void noTone(pin_size_t pinNumber) {
+  k_timer_stop(&arduino_pin_timers[pinNumber]);
+  gpio_pin_set_dt(&arduino_pins[pinNumber], 0);
+}
+
 void delay(unsigned long ms) { k_sleep(K_MSEC(ms)); }
 
 void delayMicroseconds(unsigned int us) { k_sleep(K_USEC(us)); }


samples/blinky_tone/README.rst
```diff
@@ -0,0 +1,71 @@
+.. _blinky-tone-sample:
+
+Blinky Tone
```
soburi: Although the LED is available in common,
the purpose of this API is to play a sound, so it's not a good choice.
You should add Kconfig to make it possible to select the pin.
(Since you can't specify a symbolic name in Kconfig, I think you'll need to receive a numeric number like CONFIG_TONE_OUT_DIGITAL_PIN_NUMBER and resolve it like so with UTIL_CAT(D, ...).)
I think it's best to set the default to pin D8, which is used in the Arduino example.

soburi (changes requested):

samples/blinky_tone/README.rst
Ayush1325: I think it would be better to copy https://docs.arduino.cc/built-in-examples/digital/toneMelody/ example. However, I do not have a piezo buzzer and thus cannot really test it. So, wouldn't it be better to remove the example from this PR and do that separately?

soburi: > I think it would be better to copy https://docs.arduino.cc/built-in-examples/digital/toneMelody/ example. 

https://github.com/arduino/arduino-examples/blob/main/examples/02.Digital/toneMelody/toneMelody.ino

The code is in the public domain so that you can import it with a license notice.
But even so, It is better to modify to make a pin configurable.

> However, I do not have a piezo buzzer and thus cannot really test it. So, wouldn't it be better to remove the example from this PR and do that separately?

If you can't test, you don't have the hardware, describe how to verify the code and notice it in the commit message.

samples/tone_melody/src/pitches.h
```diff
@@ -0,0 +1,93 @@
+/*************************************************
```
DhruvaG2000: Missing licenses

samples/tone_melody/src/main.cpp
```diff
@@ -0,0 +1,51 @@
...
+};
+
+void setup() {
+  // iterate over the notes of the melody:
+  for (int thisNote = 0; thisNote < 8; thisNote++) {
+
+    // to calculate the note duration, take one second divided by the note type.
+    //e.g. quarter note = 1000 / 4, eighth note = 1000/8, etc.
+    int noteDuration = 1000 / noteDurations[thisNote];
+    tone(CONFIG_TONE_OUT_DIGITAL_PIN_NUMBER, melody[thisNote], noteDuration);
```
DhruvaG2000: I honestly don't think this is the way to go. I don't want to go and look at a Kconfig option just to see which pin  is being used. A sample should be self contained and easy to read IMO.

I'd rather simply:
```
#define toneOutPin 8

tone(toneOutPin , melody[thisNote], noteDuration);
```
Or something similar.

samples/blinky_tone/README.rst
Ayush1325: @soburi What do you mean by import with license notice?

Ayush1325:
Found a piezo buzzer and tested with bcf. Also fixed the case for frequency = 0.

DhruvaG2000 (changes requested):

samples/tone_melody/src/pitches.h
DhruvaG2000: I see it got resolved now. ignore this.

Ayush1325:
@DhruvaG2000 What should I do regarding copyright and license?

samples/blinky_tone/README.rst
soburi: I think it is enough because it describes the code from where.

https://github.com/zephyrproject-rtos/gsoc-2022-arduino-core/pull/100/files#diff-d26d067fb4bf1ed2b5f8429f6cfa7b9020e323c7d9f29c4f121199f4e17506d0R20

samples/tone_melody/src/main.cpp
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (c) 2010-2011 Tom Igoe
+ * Copyright (c) 2024 Ayush Singh <ayush@beagleboard.org>
+ *
+ * SPDX-License-Identifier: CC-BY-1.0
```
soburi: This is perhaps `CC0-1.0`.

samples/tone_melody/src/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Plays a melody
+ *
+ * circuit:
+ * - 8 ohm speaker on digital pin 8
+ *
+ * created 21 Jan 2010
+ * modified 30 Aug 2011
+ * by Tom Igoe
+ *
```
soburi: Add your credit if it is modified.

cores/arduino/main.cpp
```diff
@@ -12,6 +12,7 @@ int main(void) {
   for (;;) {
     loop();
     if (arduino::serialEventRun) arduino::serialEventRun();
+    k_yield();
```
DhruvaG2000: None of you commits even talk about this change, and it feels unrelated to this PR.
k_can_yield maybe a better option, but something to discuss beyond the scope of this PR.

DhruvaG2000 (changes requested):

samples/tone_melody/src/main.cpp
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (c) 2010-2011 Tom Igoe
+ * Copyright (c) 2024 Ayush Singh <ayush@beagleboard.org>
+ *
+ * SPDX-License-Identifier: CC0-1.0
```
DhruvaG2000: This seems like a very liberal license: https://creativecommons.org/publicdomain/zero/1.0/deed.en

> You can copy, modify, distribute and perform the work, even for commercial purposes, all without asking permission. See Other Information below.
> When using or citing the work, you should not imply [endorsement ](https://creativecommons.org/publicdomain/zero/1.0/deed.en#ref-endorsement)by the author or the affirmer.


I think best to remove it from here and make this file in this context as Zephyr dictates, Apache 2. I am also okay with not having this sample at all if it creates licensing issues, we can always only point to it's link in documentation (with changes required if any).

@soburi  thoughts?

Ayush1325:
@DhruvaG2000 I think it might be best to remove the example from this PR. The upstream example can be used without any modification, so it is not particularly important to have it here, especially blocking this PR.

samples/tone_melody/src/main.cpp
soburi: License changes can only be made by the copyright holder.
In this case, since we are not the authors of this sample, we need to either inherit the original license or not use this sample.
It's not a complex sample, so it might be best to recreate it.

Ayush1325: I have dropped the sample from this PR.

DhruvaG2000 (approved):
I Am ok with this.

soburi (approved):

PR #102, opened 2024-06-08
zephyrCommon: Implement random and randomSeed

Ayush1325:
- Using zephyr random api
- Using a stub for randomSeed since it is not present in Zephyr

### DIFF
--- a/cores/arduino/zephyrCommon.cpp
+++ b/cores/arduino/zephyrCommon.cpp
@@ -328,3 +328,21 @@ void detachInterrupt(pin_size_t pinNumber)
 {
   setInterruptHandler(pinNumber, nullptr);
 }
+
+#ifndef CONFIG_MINIMAL_LIBC_RAND
+
+#include <stdlib.h>
+
+void randomSeed(unsigned long seed) {
+	srand(seed);
+}
+
+long random(long min, long max) {
+	return rand() % (max - min) + min;
+}
+
+long random(long max) {
+	return rand() % max;
+}
+
+#endif


cores/arduino/zephyrCommon.cpp
```diff
@@ -328,3 +328,23 @@ void detachInterrupt(pin_size_t pinNumber)
...
+#ifndef CONFIG_RNG_GENERATOR_CHOICE
+
+#include <zephyr/random/random.h>
+
+void randomSeed(unsigned long) {}
+
+long random(long min, long max) {
+	long temp;
+
+	sys_rand_get(&temp, sizeof(temp));
```
soburi: Since Zephyr's random numbers behave differently depending on the backend, you should use the original ANSI C rand() or an equivalent linear congruential implementation here.

soburi (changes requested):

cores/arduino/zephyrCommon.cpp
Ayush1325: @DhruvaG2000 Do you have any input regarding this? I am fine with adding ANSI C implementation if that would be better. We would probably need to document somewhere that we are not using zephyr random implementation though.

cores/arduino/zephyrCommon.cpp
```diff
@@ -328,3 +328,23 @@ void detachInterrupt(pin_size_t pinNumber)
 {
   setInterruptHandler(pinNumber, nullptr);
 }
+
+#ifndef CONFIG_RNG_GENERATOR_CHOICE
+
+#include <zephyr/random/random.h>
+
+void randomSeed(unsigned long) {}
```
DhruvaG2000: Let's take inspiration from how esp32 is doing it here:
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/WMath.cpp

cores/arduino/zephyrCommon.cpp
DhruvaG2000: > // Calling randomSeed() will force the
> // Pseudo Random generator like in

See if we can do something like the esp32 arduino core is doing.

DhruvaG2000 (changes requested):

Ayush1325:
I have switched to using zephyr stdlib `rand` and `srand`. It's present even in the minimal libc, so I do not think there is any drawback to using it.

DhruvaG2000 (approved):

soburi (approved):

PR #103, opened 2024-06-12
Add SPI Library

### LINKED ISSUE #87: Implementing SPI library
We need to implement the SPI library to keep it compatible with standard AVR Arduino.

Ayush1325:
- Tested master mode on beagleconnect freedom.
- Add master sample.
- Interrupts not implemented. Slave mode not tested.
- Fixes #87


libraries/SPI/SPI.cpp
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright (c) 2024 Ayush Singh <ayush@beagleboard.org>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "SPI.h"
+#include <zephyr/kernel.h>
+
+uint8_t SPCR = 0;
```
DhruvaG2000: No need of ``= 0`` IMO
See https://stackoverflow.com/questions/16015656/are-global-variables-always-initialized-to-zero-in-c

DhruvaG2000: Also, may want to add a comment at the top saying
```
/* Serial Peripheral Control Register */
```
As some people may even call it SPIGCR or something else. 
See https://ti.com/lit/ug/sprugp2a/sprugp2a for eg.

libraries/SPI/SPI.cpp
```diff
@@ -0,0 +1,101 @@
...
+  const struct spi_buf_set tx_buf_set = {
+      .buffers = &tx_buf,
+      .count = 1,
+  };
+  const struct spi_buf rx_buf = {.buf = &rx, .len = sizeof(rx)};
+  const struct spi_buf_set rx_buf_set = {
+      .buffers = &rx_buf,
+      .count = 1,
+  };
+  int ret;
```
DhruvaG2000: No mixed declarations please.
FYI: https://langdev.stackexchange.com/questions/3115/why-are-mixed-declarations-more-challenging-to-implement-than-forcing-all-declar

libraries/SPI/SPI.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+
```
DhruvaG2000, line 29: remove extra blank lines

samples/spi_master/README.rst
```diff
@@ -0,0 +1,9 @@
+.. _hello_world:
+
+SPI Master
```
DhruvaG2000: Please make use of more inclusive terminology.  This should be addressed throughout this PR.
Read https://www.oshwa.org/a-resolution-to-redefine-spi-signal-names/

SPI controller makes more sense.

samples/spi_master/src/app.cpp
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2024 Ayush Singh <ayush@beagleboard.org>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "SPI.h"
+#include <Arduino.h>
+
+#define SS 3
```
DhruvaG2000: Very ambiguous name, use something more verbose

DhruvaG2000 (changes requested):

samples/spi_master/src/app.cpp
Ayush1325: So, the reason for choosing `SS` is that it seems to be defined in standard arduino as the CS Pin.

libraries/SPI/SPI.cpp
Ayush1325: @DhruvaG2000 the name is taken from [Arduino SPI docs](https://docs.arduino.cc/tutorials/generic/introduction-to-the-serial-peripheral-interface/). So I think we should stick with it?

libraries/SPI/SPI.cpp
Ayush1325: @DhruvaG2000 Do you just mean moving `ret` at the top or remove mixed declaration (like C89)? That would require making everything non-const.

DhruvaG2000: just move it to the top.

libraries/SPI/SPI.cpp
DhruvaG2000: @Ayush1325  My comment wasn't on the name, I'm just saying write the full form as a comment above

samples/spi_master/src/app.cpp
DhruvaG2000: https://docs.arduino.cc/tutorials/generic/introduction-to-the-serial-peripheral-interface/

chip sel/ CS makes more sense.

DhruvaG2000: > Controller/peripheral is formerly known as master/slave. Arduino no longer supports the use of this terminology. See the table below to understand the new terminology

DhruvaG2000 (approved):
I think we'll also have to add the spi node to add support for other platforms as well, but I don't suppose you'd have the hw to test it all anyway.
So I am good with this PR, thanks for addressing all concerns!

PR #104, opened 2024-06-19
samples: Remove CONFIG_CPLUSPLUS

DhruvaG2000:
This causes the following build errors:
 /__w/gsoc-2022-arduino-core/gsoc-2022-arduino-core/Arduino-Zephyr-API/samples/i2cdemo/prj.conf:2: warning: attempt to assign the value 'y' to the undefined symbol CPLUSPLUS

Hence, remove it from all samples as it is no longer needed.


DhruvaG2000:
Merging this ASAP, because it's breaking CI. We can always drop later if any objections.

PR #110, opened 2024-08-04
variants: Rename variants.h to variant.h

soburi:
The header for Arduino board-specific settings is variant.h, not variants.h, so I modified it accordingly.

such as:
https://github.com/arduino/ArduinoCore-samd/blob/master/variants/arduino_zero/variant.h


DhruvaG2000 (approved):
Fix spelling of rename in $subject please, Rest is good.

soburi:
> Fix spelling of rename in $subject please, Rest is good.

Oops... That's Embarrassing!
Fixed it.

PR #111, opened 2024-08-04
CMakelists: Support HWMv2 structure

soburi:
Change the criteria so that not only `variants/${BOARD}` but also `variants/${BOARD}${BOARD_QUALIFIERS}` are valid as variant paths.


DhruvaG2000 (approved):
Valid change, I was wondering about this since the nrf52x got it's own kind of sub variants structure

PR #114, opened 2024-08-05
samples: Remove CONFIG_CPLUSPLUS

soburi:
The configuration of CPLUSPLUS has already been removed. Remove all of the settings.
CONFIG_ARDUINO_API enables CONFIG_CPP if the option is selected. There is no need to enable C++ manually.


DhruvaG2000 (approved):
I've been meaning to do this, else builds fail for these samples.
Thanks for doing this.

PR #119, opened 2024-08-16
Implement interrupts and noInterrupts

Ayush1325:
- Using irq_lock and irq_unlock zephyr apis.
- Not really sure where these should go. `Common.h` has the following line:
```
// interrupts() / noInterrupts() must be defined by the core
```
- Need these for the [NeoPixel library](https://github.com/adafruit/Adafruit_NeoPixel).

### DIFF
--- a/cores/arduino/zephyrCommon.cpp
+++ b/cores/arduino/zephyrCommon.cpp
@@ -175,6 +175,22 @@ size_t analog_pin_index(pin_size_t pinNumber) {
 
 #endif //CONFIG_ADC
 
+static unsigned int irq_key;
+static bool interrupts_disabled = false;
+
+void interrupts(void) {
+  if (interrupts_disabled) {
+    irq_unlock(irq_key);
+    interrupts_disabled = false;
+  }
+}
+
+void noInterrupts(void) {
+  if (!interrupts_disabled) {
+    irq_key = irq_lock();
+    interrupts_disabled = true;
+  }
+}
 }
 
 void yield(void) {
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -97,6 +97,9 @@ enum analogPins { DT_FOREACH_PROP_ELEM(DT_PATH(zephyr_user),
 
 #endif
 
+void interrupts(void);
+void noInterrupts(void);
+
 #include <variant.h>
 #ifdef __cplusplus
 #include <zephyrPrint.h>


cores/arduino/zephyrCommon.cpp
```diff
@@ -464,3 +464,18 @@ void disableInterrupt(pin_size_t pinNumber) {
     pcb->handlers[BIT(arduino_pins[pinNumber].pin)].enabled = false;
   }
 }
+
+static unsigned int irq_key = UINT_MAX;
```
soburi: Please put it into an anonymous namespace that is started from line 13,
or redeclare here.

cores/arduino/zephyrCommon.cpp
```diff
@@ -175,6 +175,20 @@ size_t analog_pin_index(pin_size_t pinNumber) {
 
 #endif //CONFIG_ADC
 
+static unsigned int irq_key = UINT_MAX;
```
DhruvaG2000: I am no big fan of `UINT_MAX` , is there an alternate approach we can use here?
AFAICS I don't see a similar implementation of lock/ unlock anywhere in zephyr, that checks against something like UINT_MAX. This feels error prone to me.

cores/arduino/zephyrCommon.cpp
```diff
@@ -175,6 +175,20 @@ size_t analog_pin_index(pin_size_t pinNumber) {
...
+void interrupts(void) {
+  if (irq_key != UINT_MAX) {
+    irq_unlock(irq_key);
+    irq_key = UINT_MAX;
+  }
+}
+
+void noInterrupts(void) {
+  if (irq_key == UINT_MAX) {
+    irq_key = irq_lock();
```
DhruvaG2000, line 190: I don't fully understand from the documentation what possible values of key are. Do you know? Maybe we can rely on those than uint max?

DhruvaG2000 (changes requested):

cores/arduino/zephyrCommon.cpp
Ayush1325: So the docs just say that the values will be arch specific. We can use a bool flag or a `std::optional` instead to keep track if the `irq_lock` was already acquired.

DhruvaG2000: @Ayush1325  if bool works, then I'd go with bool.

DhruvaG2000 (changes requested):
You have defined `interrupts()` / `noInterrupts()` only in `common.cpp`. How will anyone outside of this file ever use it? Using it in some example would be great.

These function prototypes don't exist in the Arduino Core API.
See how SAMD has done it: https://github.com/arduino/ArduinoCore-samd/blob/993398cb7a23a4e0f821a73501ae98053773165b/cores/arduino/Arduino.h#L53

Ayush1325:
> You have defined `interrupts()` / `noInterrupts()` only in `common.cpp`. How will anyone outside of this file ever use it? Using it in some example would be great.
> 
> These function prototypes don't exist in the Arduino Core API. See how SAMD has done it: https://github.com/arduino/ArduinoCore-samd/blob/993398cb7a23a4e0f821a73501ae98053773165b/cores/arduino/Arduino.h#L53

Is there a reason to use macro instead of just adding function prototype to `Arduino.h`?

DhruvaG2000:
> > You have defined `interrupts()` / `noInterrupts()` only in `common.cpp`. How will anyone outside of this file ever use it? Using it in some example would be great.
> > These function prototypes don't exist in the Arduino Core API. See how SAMD has done it: https://github.com/arduino/ArduinoCore-samd/blob/993398cb7a23a4e0f821a73501ae98053773165b/cores/arduino/Arduino.h#L53
> 
> Is there a reason to use macro instead of just adding function prototype to `Arduino.h`?

Nope, not atall. I just meant to show that they have exposed it via the `Arduino.h`. You can make it a prototype.

DhruvaG2000 (approved):
Thanks for addressing all comments.

DhruvaG2000:
Can you please rebase and force push?

PR #121, opened 2024-09-02
Arduino.h: Add digitalPinToInterrupt

Ayush1325:
- Return the same number as long as not more than number of digital pins.
- This does not seem to be present in ArduinoCoreAPI but defined for all Arduino boards (and microblocks has started using it).
- Also present in arduino docs [1]

[1]: https://www.arduino.cc/reference/en/language/functions/external-interrupts/digitalpintointerrupt/

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -100,6 +100,8 @@ enum analogPins { DT_FOREACH_PROP_ELEM(DT_PATH(zephyr_user),
 void interrupts(void);
 void noInterrupts(void);
 
+int digitalPinToInterrupt(pin_size_t pin);
+
 #include <variant.h>
 #ifdef __cplusplus
 #include <zephyrPrint.h>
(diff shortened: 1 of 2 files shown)


cores/arduino/Arduino.h
```diff
@@ -97,6 +97,11 @@ enum analogPins { DT_FOREACH_PROP_ELEM(DT_PATH(zephyr_user),
 
 #endif
 
+static inline int digitalPinToInterrupt(pin_size_t pin)
```
DhruvaG2000: I think this is a very surface level implementation we have done here, I don't think it's quite accurate? I maybe wrong.

> See [attachInterrupt()](https://www.arduino.cc/reference/en/language/functions/external-interrupts/attachinterrupt) for a full list of supported interrupt pins on all boards.

Can you not check whether the pin supports interrupt or not? The current implementation just feels like it's returning pin if it exists, nothing to do with whether it supports IRQ.

DhruvaG2000 (changes requested):

Ayush1325:
I tried putting it in the anonymous namespace, but then I got undefined reference errors. So not sure if `interrupts` and `noInterrupts` are working or not.

Note: I am not a C++ dev and have never used namespaces anything newer than C++98. So my understanding here is rather limited.

DhruvaG2000 (approved):

PR #122, opened 2024-09-05
Move interrupts and noInterrupts out of namespace

Ayush1325:
- These functions are exported from Arduino.h
- Fixes #119

### DIFF
--- a/cores/arduino/zephyrCommon.cpp
+++ b/cores/arduino/zephyrCommon.cpp
@@ -177,20 +177,6 @@ size_t analog_pin_index(pin_size_t pinNumber) {
 
 static unsigned int irq_key;
 static bool interrupts_disabled = false;
-
-void interrupts(void) {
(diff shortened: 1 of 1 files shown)


cores/arduino/zephyrCommon.cpp
```diff
@@ -480,3 +463,20 @@ void disableInterrupt(pin_size_t pinNumber) {
     pcb->handlers[BIT(arduino_pins[pinNumber].pin)].enabled = false;
   }
 }
+
+static unsigned int irq_key;
```
soburi: In C++, you should use anonymous namespace instead of static.

Ayush1325: For functions as well? Because I get undefined reference errors if I move functions into the anonymous namespace

soburi (approved):

PR #131, opened 2025-08-23
zephyrCommon: Add comment at the end of the anonymous namespace

soburi:
Added a comment because it was difficult to understand where the namespace ends.


Ayush1325 (approved):

DhruvaG2000 (approved):
You're right, this is actually also stemming from the fact that this file has just got wayy too big than it should be and deserves a refactor. I will add this in my TODO.

DhruvaG2000:
@soburi no merge commits please, please  rebase and push :)

DhruvaG2000:
manually applied, thanks.

PR #133, opened 2025-08-23
zephyrCommon: Improved `tone` implemetation

soburi:
Improved to align more closely with the standard Arduino implementation.

- Eliminated the end timer and used counting to determine the end.
- Improved to handle infinity correctly.
- Added a guard to prevent the timeout value from reaching 0.
- Set the GPIO value to 0 when the timer starts.

### DIFF
--- a/cores/arduino/zephyrCommon.cpp
+++ b/cores/arduino/zephyrCommon.cpp
@@ -212,49 +212,74 @@ PinStatus digitalRead(pin_size_t pinNumber) {
   return (gpio_pin_get_dt(&arduino_pins[pinNumber]) == 1) ? HIGH : LOW;
 }
 
-struct k_timer arduino_pin_timers[ARRAY_SIZE(arduino_pins)];
-struct k_timer arduino_pin_timers_timeout[ARRAY_SIZE(arduino_pins)];
+#ifndef MAX_TONE_PINS
+#define MAX_TONE_PINS DT_PROP_LEN(DT_PATH(zephyr_user), digital_pin_gpios)
+#endif
+
+#define TOGGLES_PER_CYCLE 2ULL
+
+static struct pin_timer {
+  struct k_timer timer;
+  uint32_t count;
+  pin_size_t pin;
+  bool infinity;
+} arduino_pin_timers[MAX_TONE_PINS];
 
 void tone_expiry_cb(struct k_timer *timer) {
-  const struct gpio_dt_spec *spec = (gpio_dt_spec*)k_timer_user_data_get(timer);
-  gpio_pin_toggle_dt(spec);
+  struct pin_timer *pt = CONTAINER_OF(timer, struct pin_timer, timer);
+  const struct gpio_dt_spec *spec = &arduino_pins[pt->pin];
+
+  if (pt->count == 0) {
+    k_timer_stop(timer);
+    gpio_pin_set_dt(spec, 0);
+  } else {
+    gpio_pin_toggle_dt(spec);
+    if (!pt->infinity) {
+      pt->count--;
+    }
+  }
 }
 
-void tone_timeout_cb(struct k_timer *timer) {
-  pin_size_t pinNumber = (pin_size_t)(uintptr_t)k_timer_user_data_get(timer);
-  noTone(pinNumber);
-}
-
-void tone(pin_size_t pinNumber, unsigned int frequency, unsigned long duration) {
-  struct k_timer *timer = &arduino_pin_timers[pinNumber];
+void tone(pin_size_t pinNumber, unsigned int frequency,
+          unsigned long duration) {
   const struct gpio_dt_spec *spec = &arduino_pins[pinNumber];
+  struct k_timer *timer;
   k_timeout_t timeout;
 
+  if (pinNumber >= MAX_TONE_PINS) {
+    return;
+  }
+
+  timer = &arduino_pin_timers[pinNumber].timer;
+
   pinMode(pinNumber, OUTPUT);
+  k_timer_stop(&arduino_pin_timers[pinNumber].timer);
 
   if (frequency == 0) {
     gpio_pin_set_dt(spec, 0);
     return;
   }
 
-  timeout = K_NSEC(NSEC_PER_SEC / (2 * frequency));
-
-  k_timer_init(timer, tone_expiry_cb, NULL);
-  k_timer_user_data_set(timer, (void*)spec);
-  gpio_pin_set_dt(spec, 1);
-  k_timer_start(timer, timeout, timeout);
-
-  if(duration > 0) {
-    timer = &arduino_pin_timers_timeout[pinNumber];
-    k_timer_init(timer, tone_timeout_cb, NULL);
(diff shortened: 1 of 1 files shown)


cores/arduino/zephyrCommon.cpp
```diff
@@ -212,49 +212,70 @@ PinStatus digitalRead(pin_size_t pinNumber) {
...
 
-void tone_timeout_cb(struct k_timer *timer) {
-  pin_size_t pinNumber = (pin_size_t)(uintptr_t)k_timer_user_data_get(timer);
-  noTone(pinNumber);
+  if (pt->count == 0) {
+    k_timer_stop(timer);
+    gpio_pin_set_dt(spec, 0);
+  } else {
+    gpio_pin_toggle_dt(spec);
+    if (pt->count != UINT32_MAX) {
```
Ayush1325: Maybe it would be better to add a `bool infinite` to `struct pin_timer`? Other than that, it looks good to me.
Using only a single timer is great, should help fix a lot of edge cases.

Ayush1325 (approved):

cores/arduino/zephyrCommon.cpp
soburi: > Maybe it would be better to add a bool infinite to struct pin_timer? 

I applied your suggestion.

cores/arduino/zephyrCommon.cpp
```diff
@@ -212,49 +212,65 @@ PinStatus digitalRead(pin_size_t pinNumber) {
...
-struct k_timer arduino_pin_timers_timeout[ARRAY_SIZE(arduino_pins)];
+#ifndef MAX_TONE_PINS
+#define MAX_TONE_PINS DT_PROP_LEN(DT_PATH(zephyr_user), digital_pin_gpios)
+#endif
+
+struct pin_timer {
+  struct k_timer timer;
+  pin_size_t pin;
+  bool infinity;
+  uint32_t count;
```
Ayush1325: I think it would be better to place `bool infinity` after count. Decreasing order of size for efficient size: https://markaicode.com/ultimate-guide-to-c-structure-memory-alignment-and-padding/

```c
struct pin_timer {
  struct k_timer timer;
  uint32_t count;
  pin_size_t pin;
  bool infinity;
} arduino_pin_timers[MAX_TONE_PINS];
```

Also moved `pin` down since it can either be 32bit (=count) or 8 bit (<count).

Ayush1325 (approved):

cores/arduino/zephyrCommon.cpp
```diff
@@ -212,49 +212,65 @@ PinStatus digitalRead(pin_size_t pinNumber) {
...
     return;
   }
 
   timeout = K_NSEC(NSEC_PER_SEC / (2 * frequency));
+  if (timeout.ticks == 0) {
+    timeout.ticks = 1;
+  }
 
+  arduino_pin_timers[pinNumber].infinity = (duration == 0);
+  arduino_pin_timers[pinNumber].count = ((uint64_t)duration * frequency) / 500ULL;
```
DhruvaG2000: Please avoid usage of magic numbers, can you call out why this is `/ 500`?

cores/arduino/zephyrCommon.cpp
```diff
@@ -212,49 +212,65 @@ PinStatus digitalRead(pin_size_t pinNumber) {
   return (gpio_pin_get_dt(&arduino_pins[pinNumber]) == 1) ? HIGH : LOW;
 }
 
-struct k_timer arduino_pin_timers[ARRAY_SIZE(arduino_pins)];
-struct k_timer arduino_pin_timers_timeout[ARRAY_SIZE(arduino_pins)];
+#ifndef MAX_TONE_PINS
```
DhruvaG2000, line 215: $commit title: fix  "implemetation" → "implementation"

cores/arduino/zephyrCommon.cpp
```diff
@@ -212,49 +212,65 @@ PinStatus digitalRead(pin_size_t pinNumber) {
...
+#define MAX_TONE_PINS DT_PROP_LEN(DT_PATH(zephyr_user), digital_pin_gpios)
+#endif
+
+struct pin_timer {
+  struct k_timer timer;
+  uint32_t count;
+  pin_size_t pin;
+  bool infinity;
+} arduino_pin_timers[MAX_TONE_PINS];
```
DhruvaG2000, line 226: Consider adding explicit initialization `arduino_pin_timers[MAX_TONE_PINS] = {}; `

cores/arduino/zephyrCommon.cpp
```diff
@@ -212,49 +212,65 @@ PinStatus digitalRead(pin_size_t pinNumber) {
...
+  }
 }
 
-void tone(pin_size_t pinNumber, unsigned int frequency, unsigned long duration) {
-  struct k_timer *timer = &arduino_pin_timers[pinNumber];
+void tone(pin_size_t pinNumber, unsigned int frequency,
+          unsigned long duration) {
   const struct gpio_dt_spec *spec = &arduino_pins[pinNumber];
+  struct k_timer *timer = &arduino_pin_timers[pinNumber].timer;
   k_timeout_t timeout;
```
DhruvaG2000, line 247: Add bounds checking, 
```c
if (pinNumber >= MAX_TONE_PINS) return; 
```

cores/arduino/zephyrCommon.cpp
DhruvaG2000: Something like this helps make it more readable
```c
#define MS_PER_SECOND 1000ULL                                                                                                                                              
#define TOGGLES_PER_CYCLE 2ULL
#define TONE_COUNT_DIVISOR (MS_PER_SECOND / TOGGLES_PER_CYCLE)  // = 500
```

DhruvaG2000 (changes requested):
Also $subject should be imperative -> `zephyrCommon: Improve tone implemetation`

cores/arduino/zephyrCommon.cpp
soburi: Addressed.

cores/arduino/zephyrCommon.cpp
soburi: Addressed.

cores/arduino/zephyrCommon.cpp
soburi: Added this check.

cores/arduino/zephyrCommon.cpp
soburi: This is not necessary because the static region ensures initialization to 0 according to the language specification.

DhruvaG2000 (approved):
Sigh... all the CI failures seem to be github/ infra related :( 
Ran checkpatch locally and saw no errors. So approving.
@soburi  assuming you've had a chance to build test locally?

soburi:
> Sigh... all the CI failures seem to be github/ infra related :(
> Ran checkpatch locally and saw no errors. So approving.
> @soburi assuming you've had a chance to build test locally?

Yes, I verified build and run with my arduino_mkrzero.

PR #134, opened 2025-08-23
[DNM] zephyrCommon: Changing pin numbering rules

soburi:
Zephyr's [ba48d83b](https://github.com/zephyrproject-rtos/zephyr/commit/ba48d83bec18e9f2caf61979aadf9f8537086cee)
changes make it easy to configure pin numbering at compile,
so we'll change the pin numbers to be based on the GPIO pin numbers.

This is same as the traditional Arduino, using the GPIO numbers as
pin numbers. Pin names written on the board, such as D1 and D2,
are aliases for these pin numbers.

When using multiple GPIO ports, pin numbers are numbered consecutively
from the beginning. For example, if you have two 16-port GPIOs,
16 refers to the first pin (idx=0) of the second port.

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2022 Dhruva Gole
+ * Copyright (c) 2025 TOKITA Hiroshi
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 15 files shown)


pillo79:
Just chiming in for the Arduino side to say, please don't drop custom pin numbering. 

The technical implementation is great :+1: and it removes the need for bulky pin lists :yellow_heart:, but AFAICS also makes it "impossible" to assign _specific_ numbers to pins, and this is something that is really important for backwards compatibility with traditional Arduino sketches.

[ OTOH, I do hate having to write those lists; tried several alternative DTS mappings myself but haven't yet found one that is flexible enough and/or enough of an improvement... :slightly_frowning_face: ]

soburi:
> Just chiming in for the Arduino side to say, please don't drop custom pin numbering.
> 
> The technical implementation is great 👍 and it removes the need for bulky pin lists 💛, but AFAICS also makes it "impossible" to assign _specific_ numbers to pins, and this is something that is really important for backwards compatibility with traditional Arduino sketches.
> 
> [ OTOH, I do hate having to write those lists; tried several alternative DTS mappings myself but haven't yet found one that is flexible enough and/or enough of an improvement... 🙁 ]

The main purpose of this change is to make pins that are not exposed on the connector usable.
The pin change itself is necessary to deal with edge cases, so it needs to remain.
My current idea is to use the https://github.com/zephyrproject-rtos/zephyr/pull/87595 change to extract pin information from the gpio-nexus node that is labeled as `arduino_header`.
With this method, as long as the connector is defined on the devicetree, I think it will be possible to make it behave like an "Arduino" without writing any additional config. For edge cases that still cannot be solved, I think we'll have to deal with them with the current settings, but some consideration may be needed to find the best way to handle them.

---
By the way, https://github.com/zephyrproject-rtos/zephyr/pull/87595 is stalled because only a few people reviewed it.
If you can review this part, please help with this PR review.

soburi:
@pillo79 

> The technical implementation is great 👍 and it removes the need for bulky pin lists 💛, but AFAICS also makes it "impossible" to assign specific numbers to pins, and this is something that is really important for backwards compatibility with traditional Arduino sketches.

Does this mean you want to freely map the direct values ​​passed to digitalOut?

I understand that traditionally, Arduino's D1-based values ​​were logical pin numbers,
while values ​​like digitalOut argument were physical pin numbers (i.e., implementation-dependent).

Until now, the gsoc-2022-arduino-core implementation has assumed logical = physical due to various constraints. (i.e., D0 = 0, D1 = 1)

In this implementation, only mapped pins are usable, so the intention of this PR is to introduce physical pin numbers to handle all pins on the board.

If the answer to the initial question is yes, I think this can be addressed by adding one more LUT.

cores/arduino/zephyrCommon.cpp
```diff
@@ -1,30 +1,71 @@
...
+                                  : local_gpio_port_r(pin, ctrl + 1, accum + end[0], end + 1, n - 1);
 }
 
-template <class N, class Head, class... Tail>
-constexpr const N sum_of_list(const N sum, const Head &head, const Tail &...tail)
-{
-  return sum_of_list(sum + head, tail...);
+constexpr inline size_t port_index_r(const struct device *target,
+                                     const struct device *const *table, pin_size_t idx, size_t n) {
+  return (n == 0)               ? size_t(-1)
```
DhruvaG2000, line 34: Is it me or is the alignment really messed up on this one?

DhruvaG2000 (commented):
I have seen the PR, and have some catching up to do on the new zephyr commit that has been mentioned. I will need some more time to actually go through since this seems to have a large impact overall. Please excuse the delay!

cores/arduino/Arduino.h
```diff
@@ -103,6 +48,11 @@ void noInterrupts(void);
 int digitalPinToInterrupt(pin_size_t pin);
 
 #include <variant.h>
+
+#if !defined(LED_BUILTIN) && DT_NODE_EXISTS(DT_ALIAS(led0))
+#define LED_BUILTIN ZARD_GLOBAL_GPIO_NUM(DT_ALIAS(led0), gpios, 0)
+#endif // LED_BUILTIN
+
```
DhruvaG2000, line 55: No LED_BUILTIN definition if variant.h doesn't define it AND there's no led0 alias
Can we also have add a static assert to catch this at compile time:
```
#if !defined(LED_BUILTIN) && !DT_NODE_EXISTS(DT_ALIAS(led0))
#error "Board has no LED_BUILTIN defined and no led0 alias. Define LED_BUILTIN in variant.h."
#endif                                                                                                                             
```

cores/arduino/zephyrCommon.cpp
```diff
@@ -1,30 +1,71 @@
...
+  DT_FOREACH_PROP_ELEM_SEP(DT_PATH(zephyr_user), gpio_ports, PROP_NGPIOS, (,))
+};
+
+constexpr inline const struct device *local_gpio_port_r(pin_size_t pin,
+                                                        const struct device *const *ctrl,
+                                                        const uint32_t accum,
+							const uint32_t *end, size_t n) {
+  return (n == 0)                 ? nullptr
+         : (pin < accum + end[0]) ? ctrl[0]
+                                  : local_gpio_port_r(pin, ctrl + 1, accum + end[0], end + 1, n - 1);
```
DhruvaG2000, line 29: local_gpio_port_r returns nullptr but calling code doesn't check.
Please add null checks in caller

cores/arduino/zephyrCommon.cpp
```diff
@@ -1,30 +1,71 @@
...
+constexpr inline pin_size_t end_accum_r(const uint32_t accum, const uint32_t *end, size_t n) {
+  return (n == 0) ? accum : end_accum_r(accum + end[0], end + 1, n - 1);
+}
+
+constexpr inline pin_size_t end_accum(size_t n) {
+  return end_accum_r(0, pins, n);
+}
+
+constexpr inline pin_size_t local_gpio_pin(pin_size_t gpin) {
+  return port_idx(gpin) == pin_size_t(-1) ? pin_size_t(-1) : gpin - end_accum(port_idx(gpin));
```
DhruvaG2000, line 56: Maybe better to do it this way:
```
const struct device *port = local_gpio_port(gpin);
if (port == nullptr) {
return PIN_INVALID // or some #define for invalid pins
}
```

DhruvaG2000: Then update all GPIO functions to check for PIN_INVALID

DhruvaG2000 (changes requested):
@Ayush1325  if possible can you test for any regressions on the Beagleconnect freedom? 
I am ok with the general idea here, and it does seem like an improvement for handling gpio ports.
Just need to be careful about the code though, and hence giving some comments from maintenance and debugging perspective.

cores/arduino/Arduino.h
soburi: This will make it impossible to use boards that do not have LEDs, so it is best not to do this.

cores/arduino/zephyrCommon.cpp
soburi: > Maybe better to do it this way:
> 
> ```
> const struct device *port = local_gpio_port(gpin);
> if (port == nullptr) {
> return PIN_INVALID // or some #define for invalid pins
> }
> ```
> 

This is a constexpr function, so conditional statements cannot be used.

cores/arduino/Arduino.h
DhruvaG2000: ok

cores/arduino/zephyrCommon.cpp
DhruvaG2000: > > Maybe better to do it this way:
> > ```
> > const struct device *port = local_gpio_port(gpin);
> > if (port == nullptr) {
> > return PIN_INVALID // or some #define for invalid pins
> > }
> > ```
> 
> This is a constexpr function, so conditional statements cannot be used.

I think it's allowed since C++14. But fine...

DhruvaG2000 (approved):

cores/arduino/zephyrCommon.cpp
soburi: Yes, C++14 supports this semanitcs, but the zephyr default is C++11.

DhruvaG2000:
I'll merge this as soon as I am done fixing the CI failures. Please excuse the delay

Ayush1325:
> @Ayush1325 if possible can you test for any regressions on the Beagleconnect freedom? I am ok with the general idea here, and it does seem like an improvement for handling gpio ports. Just need to be careful about the code though, and hence giving some comments from maintenance and debugging perspective.

Apologies for the late reply. Seems to be working fine on bcf.

pillo79:
@soburi, apologies for getting back so late, was derailed by the linked PR and most importantly all of this week's chaos. :sweat_smile: 

> Does this mean you want to freely map the direct values ​​passed to digitalOut?
> 
> I understand that traditionally, Arduino's D1-based values ​​were logical pin numbers, while values ​​like digitalOut argument were physical pin numbers (i.e., implementation-dependent).

In fact, and very unfortunately, the "official" way of referring to pins is by number. Some cores don't even define `Dx` symbols; all samples use integers to access pins, so do 99% of sketches out there...  for existing boards, numbers are the standard we need to maintain compatibility with. 

> In [the current] implementation, only mapped pins are usable, so the intention of this PR is to introduce physical pin numbers to handle all pins on the board.

Understood; with this change you assign a number for any "GPIO bit" (even if unconnected or not even exposed by the SOC!), and then map `Dx=x`.

> If the answer to the initial question is yes, I think this can be addressed by adding one more LUT.

My issue with a LUT is that it will create two sets of _indistinguishable_ integers : the "logical" ones (inputs to the LUT) and the "physical" ones (after the LUT is applied). 

Now, :drum:, when a sketch then calls `digitalWrite(4, HIGH)`, does `4` refer to a logical or physical pin?
Surely it must be logical (otherwise, what's the point of the LUT?), but it must also be physical (otherwise, the objective of this PR is broken!). :question: :face_with_diagonal_mouth: 

We have been bitten HARD by this with the Arduino Nano ESP32 - the GPIO controller there already uses numbers, but we need to map them to keep "standard Nano" pin layout. We added a LUT so that APIs would input logical and address physical. Chaos ensued, and 2 years later this is still a raw topic on the Arduino Forums.

----
(brainstorming)

There MAY be a way out. The default index pin type `pin_size_t` is 8-bit unsigned, and very few targets exceed that (but we do have a 32-bit option by [defining `EXTENDED_PIN_MODE`](https://github.com/arduino/ArduinoCore-API/blob/cd91833d90b4fe50e428021ba5051e2b7ceafc84/api/Common.h#L108-L113)). _What if_ you enabled that and made your physical mapping in some high enough range? Numbers below, say, 32k would be logical, the others physical.

* All APIs with pin numbers would accept an `uint32_t` and, if it's logical, apply the LUT to get a physical index.
* As long as you use `int` or the more correct `pin_size_t` to store a pin number, you have no problems.
* Existing sketches that use `uint8_t` work fine on existing boards, because numbers < 256 are definitely logical and thus mapped by the LUT.
* The only issue is if you want to use an existing sketch on previously unmapped physical pins, but then you're breaking new ground, might as well fix the sketch. 

(all of this needs a constant LUT, probably magically generated from the DT in some way. Would the recent `map`s help?)

Let me know what you guys think!

soburi:
@pillo79 

Sorry for my late response.

> @soburi, apologies for getting back so late, was derailed by the linked PR and most importantly all of this week's chaos. 😅

I know it was a difficult situation. You must be tired.

> In fact, and very unfortunately, the "official" way of referring to pins is by number. Some cores don't even define `Dx` symbols; all samples use integers to access pins, so do 99% of sketches out there... for existing boards, numbers are the standard we need to maintain compatibility with.

I understood what that means is  "all is device dependent'.
And there is a requirement to make it compatible with it.

> My issue with a LUT is that it will create two sets of _indistinguishable_ integers : the "logical" ones (inputs to the LUT) and the "physical" ones (after the LUT is applied).
> 
> Now, 🥁, when a sketch then calls `digitalWrite(4, HIGH)`, does `4` refer to a logical or physical pin? Surely it must be logical (otherwise, what's the point of the LUT?), but it must also be physical (otherwise, the objective of this PR is broken!). ❓ 🫤


In my opinion, they are distinct clearly.

First, there is a need to clarify the terminology.

physical pin number: The number statically derived from the hardware configuration. The 'serialized GPIO pin number' in the my proposal can be derived from the GPIO port and the pin number, so it is a type of "physical pin number".

Logical Pin Number = The number defined by the rule.
Usually, that is ruled by the API specification.
(It is not required in the term definition, but) and it is expected to map to the physical pin number without contradiction.
Of course, "use as is" rule can also be applied in this context.

digitalWrite() (And all other APIs) must take a logical pin number.
The physical pin number is derived via LUT from the logical pin number.
In other words, the physical pin number is just an internal representation. and the logical pin number is an alias for it.
All API must use the logical pin numbering system.

If we keep this principle, it will not cause a serious problem.


> We have been bitten HARD by this with the Arduino Nano ESP32 - the GPIO controller there already uses numbers, but we need to map them to keep "standard Nano" pin layout. We added a LUT so that APIs would input logical and address physical. Chaos ensued, and 2 years later this is still a raw topic on the Arduino Forums.

I couldn't understand the details, I need to check a little more carefully.
But as far as I understood,
This seems to be a problem caused by two different numbering systems existing concurrently.

1. The API numbering system should always be based on a single rule.

This is the principal.

2. In situations where this rule is not followed, individual, topical　handling should be taken.

On that basis,

3. It is possible to create comprehensive pin mapping.


> (brainstorming)
> 
> There MAY be a way out. The default index pin type `pin_size_t` is 8-bit unsigned, and very few targets exceed that (but we do have a 32-bit option by [defining `EXTENDED_PIN_MODE`](https://github.com/arduino/ArduinoCore-API/blob/cd91833d90b4fe50e428021ba5051e2b7ceafc84/api/Common.h#L108-L113)). _What if_ you enabled that and made your physical mapping in some high enough range? Numbers below, say, 32k would be logical, the others physical.

Zephyr's pin numbers are specified in 32 bits, so I don't think there's any problem with this policy.
I also think it's a good idea to make EXTENDED_PIN_MODE the default.

PR #135, opened 2025-08-23
CI: Fix failure

soburi:
CI is failing for several reasons.

- The cmsis_6 module is now required for the build.
- The board name needs to be a canonical name.

We will fix these.

Also, west.yml references the main repository,
not the one in the PR, so we will change it to use the one in the PR.
This workflow uses the Linux version of `checkpatch.pl`.
We change it to zephyr's version.


soburi:
@DhruvaG2000 @Ayush1325

Could you take a look, please?

.github/workflows/license_check.yml
Ayush1325, line 1: Can you explain why the changes to license_check are required?

soburi: The github action is stuck with the current version.
It was originally brought over from the Zephyr main body, so I applied the current one.

Updated commit message.

Ayush1325 (approved):

soburi:
@DhruvaG2000 Could you take a look, please?

DhruvaG2000 (approved):
Alot has changed of recent in zephyr seems like.
Thanks for fixing CI!

PR #138, opened 2025-10-07
ci: fix space issues in docker

DhruvaG2000:
github actions keep failing due to errors like these:

failed to register layer: write /opt/toolchains/zephyr-sdk-0.17.4/xtensa-nxp_rt700_hifi4_zephyr-elf/libexec/gcc/xtensa-nxp_rt700_hifi4_zephyr-elf/12.2.0/cc1: no space left on device
Error: Docker pull failed with exit code 1

Fix this by freeing up disk space on the runner before the job runs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,39 +4,53 @@ on: [push, pull_request]
 
 jobs:
   build:
+    name: Build Arduino-API Zephyr samples
     runs-on: ubuntu-latest
-    container: zephyrprojectrtos/ci:latest
+    container: zephyrprojectrtos/ci-base:latest
     env:
       CMAKE_PREFIX_PATH: /opt/toolchains
+      CCACHE_IGNOREOPTIONS: -specs=*
+      MODULE_PATH: ../modules/lib/Arduino-Zephyr-API
+
     steps:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          path: Arduino-Zephyr-API
+          fetch-depth: 0
+          persist-credentials: false
+          path: subfolder
+
+      - name: Fix module path, list needed HALs
+        run: |
+          mkdir -p $(dirname $MODULE_PATH) && mv subfolder $MODULE_PATH
+
+      - name: Setup Zephyr project
+        uses: zephyrproject-rtos/action-zephyr-setup@v1
+        with:
+          toolchains: arm-zephyr-eabi
+          manifest-file-name: ${{ env.MODULE_PATH }}/west.yml
+          enable-ccache: false
+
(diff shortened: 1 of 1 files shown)


DhruvaG2000:
@pillo79  will clean up the commits later, but I have borrowed a few stuff from the arduino fork here, while also makig sure it still works with the way this repo is... hope to get your reviews!

.github/workflows/build.yml
```diff
@@ -3,40 +3,69 @@ name: Build
...
+
+      - name: Fix module path, list needed HALs
+        run: |
+          mkdir -p $(dirname $MODULE_PATH) && mv subfolder $MODULE_PATH
+          NEEDED_HALS=$(grep 'build.zephyr_hals=' $MODULE_PATH/boards.txt | cut -d '=' -f 2 | xargs -n 1 echo | sort -u)
+          HAL_FILTER="-hal_.*"
+          for hal in $NEEDED_HALS; do
+            HAL_FILTER="$HAL_FILTER,+$hal"
+          done
+          echo "HAL_FILTER=$HAL_FILTER" | tee -a $GITHUB_ENV
```
pillo79: Note that this depends on `boards.txt`, which is Arduino packaging stuff. For the module test I think it would be better to hardcode the minimal list of HALs you need for the tests below (or  just download everything, this is just a CI time saver 🙂).

.github/workflows/build.yml
```diff
@@ -3,40 +3,69 @@ name: Build
 on: [push, pull_request]
 
 jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Cleanup Docker images and containers
+        run: |
+          docker system prune -af
+          docker volume prune -f
```
pillo79: Curious, what do these do? I thought the GH runner started pretty clean?

pillo79 (commented):
LGTM, minor remarks!

.github/workflows/build.yml
DhruvaG2000: Hmm let me try removing these, just something that github copilot suggested as the first fix to try :)

.github/workflows/build.yml
DhruvaG2000: @pillo79 thanks for the review. Yes I will remove this chunk for now, I don't think minimal HAL is required since I mostly commented out the filter part and it is working fine without it too.
Let's add it back later if we feel like we're running out of space again. How does that sound?

.github/workflows/build.yml
DhruvaG2000: removed.

pillo79 (approved):

PR #141, opened 2025-11-02
rust: Add a Rust compatible API implementation.

soburi:
We will gradually add a Rust compatible API implementation conforming to Apache-2.
This PR adds an alternative implementation of Common.cpp.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,6 +40,15 @@ jobs:
           git clone https://github.com/arduino/ArduinoCore-API.git $MODULE_PATH/../ArduinoCore-API
           cp -rfp $MODULE_PATH/../ArduinoCore-API/api $MODULE_PATH/cores/arduino/
 
+      - name: Setup Rust toolchain
+        run: |
+          wget https://sh.rustup.rs
+          mv index.html rustup_install.sh
+          bash rustup_install.sh -y
+          /root/.cargo/bin/rustup default stable
+          /root/.cargo/bin/rustup target add thumbv6m-none-eabi thumbv7em-none-eabihf thumbv7em-none-eabi thumbv7m-none-eabi
+          echo "/root/.cargo/bin" >> $GITHUB_PATH
+
       - name: Build fade
         run: |
           west build -p -b arduino_nano_33_ble/nrf52840/sense $MODULE_PATH/samples/fade
--- a/Kconfig
+++ b/Kconfig
@@ -17,6 +17,10 @@ config ARDUINO_API
 
 if ARDUINO_API
 
+config USE_ARDUINO_API_RUST_IMPLEMENTATION
+	bool "Use Rust implementation API core"
+	select RUST
+
 config QEMU_ICOUNT
 	bool "QEMU icount mode"
 	default n
--- a/rust/src/common.rs
+++ b/rust/src/common.rs
@@ -1,0 +1,22 @@
+// Copyright (c) 2025 TOKITA Hiroshi
+// SPDX-License-Identifier: Apache-2.0
+
+#[unsafe(no_mangle)]
+pub extern "C" fn map_i32(
+    x: i32, in_min: i32, in_max: i32, out_min: i32, out_max: i32
+) -> i32 {
+    let num = x.wrapping_sub(in_min).wrapping_mul(out_max.wrapping_sub(out_min));
+    let den = in_max.wrapping_sub(in_min);
+    // Note: To keep compatibility, the panic when den=0 is left as is.
+    num / den.wrapping_add(out_min)
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn makeWord_w(w: u16) -> u16 {
+    w
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn makeWord_hl(h: u8, l: u8) -> u16 {
+    ((h as u16) << 8) | (l as u16)
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+rust/Cargo.lock
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,4 +10,76 @@ if (CONFIG_ARDUINO_API)
   add_subdirectory(cores)
   add_subdirectory(libraries)
   zephyr_include_directories(${variant_dir})
+
+  if (CONFIG_USE_ARDUINO_API_RUST_IMPLEMENTATION)
+    # quote from https://github.com/zephyrproject-rtos/zephyr-lang-rust/blob/main/CMakeLists.txt
+    function (rust_target_arch RUST_TARGET)
+      # Map Zephyr targets to LLVM targets.
+      if(CONFIG_CPU_CORTEX_M)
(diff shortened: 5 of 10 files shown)


Kconfig
```diff
@@ -13,6 +13,7 @@ config ARDUINO_API
 	imply CBPRINTF_FP_SUPPORT
 	imply RING_BUFFER
 	select UART_INTERRUPT_DRIVEN
+	select RUST
```
pillo79: I think this should currently be an optional feature, to avoid adding the strong dependency right away.

pillo79 (commented):
Awesome, looking forward to see this grow! 🥳 🤗

.github/workflows/build.yml
```diff
@@ -39,6 +39,12 @@ jobs:
         run: |
           git clone https://github.com/arduino/ArduinoCore-API.git $MODULE_PATH/../ArduinoCore-API
           cp -rfp $MODULE_PATH/../ArduinoCore-API/api $MODULE_PATH/cores/arduino/
+          apt install rustup
```
DhruvaG2000: The recommended way to install Rust is via the rustup.rs installer script, Ubuntu's rustup package from apt is often outdated and may not work correctly

.github/workflows/build.yml
```diff
@@ -39,6 +39,12 @@ jobs:
...
+          rustup default stable
+          rustup target add thumbv6m-none-eabi
+          rustup target add thumbv7em-none-eabihf
+          rustup target add thumbv7em-none-eabi
+          rustup target add thumbv7m-none-eabi
```
DhruvaG2000: Installing targets one-by-one is slower. Rustup supports installing multiple targets in one command

DhruvaG2000:
```suggestion
          rustup target add thumbv6m-none-eabi thumbv7em-none-eabihf thumbv7em-none-eabi thumbv7m-none-eabi
```

.github/workflows/build.yml
DhruvaG2000: Also,
All initialization is lumped into one "Initialize" step mixing unrelated concerns (cloning API, copying files, Rust setup)...
  Better structure:
```
  - name: Clone ArduinoCore-API
[...]
 - name: Install Rust toolchain
[...]
```

rust/src/common.rs
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#[unsafe(no_mangle)]
+pub extern "C" fn map_i32(
+    x: i32, in_min: i32, in_max: i32, out_min: i32, out_max: i32
+) -> i32 {
+    let num = x.wrapping_sub(in_min).wrapping_mul(out_max.wrapping_sub(out_min));
+    let den = in_max.wrapping_sub(in_min);
+    // Note: To keep compatibility, the panic when den=0 is left as is.
+    num / den.wrapping_add(out_min)
```
DhruvaG2000, line 11: Due to operator precedence, this evaluates as:             
```                                                                                                                                                      
  num / (den + out_min)  // Division happens AFTER adding out_min to den!                                                                                                                                          
```
Original C++ code:
```cpp
  return (x - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
```

Can we just follow similar arithmetic syntax in rust too, the code using `wrapping_*` is making the logic really confusing to understand, and with rust already being a new addition I feel like we should make an effort to keep things simple where possible.

DhruvaG2000 (changes requested):

Kconfig
```diff
@@ -17,6 +17,11 @@ config ARDUINO_API
 
 if ARDUINO_API
 
+config USE_ARDUINO_API_RUST_IMPLEMENTATION
+	bool "Use Rust implementation API core"
+	default y
```
pillo79: Sorry to be nitpicky, but I asked this before and I'm still confused. 🙂 

Is Rust expected to be installed by anyone trying to use the core at this stage?
If not, I would expect this to be manually set somehow (via the west command line, using snippets, etc) and not being `default y`-ed.

DhruvaG2000: +1 @pillo79 
I would prefer this be optional and not default y too.

.github/workflows/build.yml
soburi: Fixed. Thank you for your comment.

.github/workflows/build.yml
soburi: Fixed.

Kconfig
soburi: OK. This is still in an experimental state.
Removed the `default y`.

rust/src/common.rs
soburi: Since this is not a "new feature" but a reimplementation of the original implementation, we believe compatibility should be maintained.

From another perspective, this implementation is an internal implementation of the C++ API,
and therefore does not take advantage of Rust's features.
To take advantage of Rust's features would require a complete overhaul of the API set, so we think it's better to leave it this way for now.
It would also be good to have an option for more Rust-like behavior in the future.

.github/workflows/build.yml
DhruvaG2000: @soburi  you didn't address the first comment, why are we not using `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh` instead of apt?

rust/src/common.rs
DhruvaG2000: ok it seems like there were some debates on how the behaviour shoul look like on the original Arduino API:
https://github.com/arduino/Arduino/issues/10382
and
https://github.com/arduino/ArduinoCore-API/issues/51

I will have to read up a bit to understand what's going on with map() in arduino

.github/workflows/build.yml
soburi: I missed it. It was addressed.

rust/src/common.rs
DhruvaG2000: Will get back to this at a later point, anyway no objections for this PR

DhruvaG2000 (approved):

DhruvaG2000:
@pillo79  can you please take another look if you still have pending comments?

Kconfig
```diff
@@ -17,6 +17,10 @@ config ARDUINO_API
 
 if ARDUINO_API
 
+config USE_ARDUINO_API_RUST_IMPLEMENTATION
```
pillo79, line 20: I am missing where this symbol is effectively used? AFAICS  CMakeLists.txt will try to invoke `cargo build` regardless of this being set or not.

pillo79 (commented):
Sorry, I might be missing something huge, but... it still seems to me that effectively Rust is being forced on every core use. What is gating the new `add_custom_command` in the main CMakeLists.txt?

Again, I think it's awesome to see this working 😍, but IMO it should be optional at this stage.

Kconfig
soburi: Oops, I missed that.
Adding if () ...

soburi:
Updated.

Could you review again? > @DhruvaG2000 @pillo79

pillo79 (approved):

soburi:
@DhruvaG2000 
Could you take a look, please?

DhruvaG2000 (approved):

PR #142, opened 2025-11-20
variant: Add PocketBeagle 2 A53 rev A1 support

Ayush1325:
- Only pins that are labeled as GPIO in schematics (and LED pins) are exposed as digital-pin-gpios. Other pins technically can be used as GPIOs, but require pinmuxing, and thus should not be part of the default setup.

### DIFF
--- a/variants/pocketbeagle_2_am6254_a53/pocketbeagle_2_am6254_a53.overlay
+++ b/variants/pocketbeagle_2_am6254_a53/pocketbeagle_2_am6254_a53.overlay
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2025 Ayush Singh <ayush@beagleboard.org>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/ {
+	zephyr,user {
+		digital-pin-gpios = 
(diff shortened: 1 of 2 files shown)


DhruvaG2000 (approved):
@Ayush1325  just remove the unnecessary `-` from the start of the commit message,
Also curious, how have you tested this? Is it with the regular u-boot then jump to zephyr binary flow?
This is exciting to have arduino support on pb-2 and I am looking fwd to test it myself as well.

Ayush1325:
> @Ayush1325 just remove the unnecessary `-` from the start of the commit message, Also curious, how have you tested this? Is it with the regular u-boot then jump to zephyr binary flow? This is exciting to have arduino support on pb-2 and I am looking fwd to test it myself as well.

Fixed the commit message.

I have tested it in 2 ways: SD Card and DFU.
The instructions for DFU are here: https://www.beagleboard.org/blog/2025-11-14-pocketbeagle-2-zephyr-dfu
The instructions for SD Card are here: https://docs.zephyrproject.org/latest/boards/beagle/pocketbeagle_2/doc/index.html#a53-core

PR #143, opened 2025-12-20
workflows: Add clang format check workflow.

soburi:
Cherry-pick the commit adding .clang-format from https://github.com/arduino/ArduinoCore-zephyr.

### DIFF
--- a/.github/workflows/format_check.yml
+++ b/.github/workflows/format_check.yml
@@ -1,0 +1,62 @@
+name: 'Format Check'
+
+on:
+  push:
+    branches:
+      - 'main'
+    paths:
+      - '**/*.c'
+      - '**/*.cpp'
+      - '**/*.h'
+      - '**/*.hpp'
+
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+    branches:
+      - 'main'
+    paths:
+      - '**/*.c'
+      - '**/*.cpp'
+      - '**/*.h'
+      - '**/*.hpp'
+
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'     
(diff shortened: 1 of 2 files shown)


.github/workflows/format_check.yml
```diff
@@ -0,0 +1,62 @@
...
+jobs:
+  format-check:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        path:
+          - check: 'cores/arduino/'
+            exclude: 'cores/arduino/api/'
+          - check: 'loader/'
+            exclude: 'loader/llext_exports\.c$'
```
DhruvaG2000, line 43: These folders don't exist in this project, cleanup to remove irrelevant stuff.

DhruvaG2000 (changes requested):
I know this is just a cherrypick, but it would still be better if you added your SoB, since you are the one cherrypicking it to this project.
Also amend the commit body to contain something useful and resolve the github copilot review issues.
Thanks!

.github/workflows/format_check.yml
soburi: I think this will be added as the integration with the Arduino repository progresses, so I would like to leave it as it is as long as it doesn't cause any errors.

soburi:
@DhruvaG2000 
Thank you for your review.

> I know this is just a cherrypick, but it would still be better if you added your SoB, since you are the one cherrypicking it to this project.

I added sign-off-by line.

> Also amend the commit body to contain something useful and resolve the github copilot review issues.
> Thanks!

Copilot's point is about formatting.
If it doesn't cause any problems, I think it would be better to reduce the difference with the arduino branch for now.
I think it is not cause a issue.

DhruvaG2000 (approved):

PR #147, opened 2026-01-31
Arduino.h: Correct LED_BUILTIN define order

soburi:
Correct the order of LED_BUILTIN definitions so that board-specific
definitions of LED_BUILTIN take precedence.

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -33,8 +33,6 @@
 
 #undef DIGITAL_PIN_CHECK_UNIQUE
 
-#ifndef LED_BUILTIN
-
 /* Return the index of it if matched, oterwise return 0 */
 #define LED_BUILTIN_INDEX_BY_REG_AND_PINNUM(n, p, i, dev, num)                                     \
 	(DIGITAL_PIN_EXISTS(n, p, i, dev, num) ? i : 0)
@@ -53,9 +51,9 @@
 	     DT_PHA_BY_IDX(DT_PATH(zephyr_user), builtin_led_gpios, 0, pin)) > 0)
 #warning "pin not found in digital_pin_gpios"
 #else
-#define LED_BUILTIN                                                                                \
-	DIGITAL_PIN_GPIOS_FIND_PIN(                                                              \
-		DT_REG_ADDR(DT_PHANDLE_BY_IDX(DT_PATH(zephyr_user), builtin_led_gpios, 0)),        \
+#define ZARD_LED_BUILTIN                                                                           \
+	DIGITAL_PIN_GPIOS_FIND_PIN(                                                                    \
+		DT_REG_ADDR(DT_PHANDLE_BY_IDX(DT_PATH(zephyr_user), builtin_led_gpios, 0)),                \
 		DT_PHA_BY_IDX(DT_PATH(zephyr_user), builtin_led_gpios, 0, pin))
 #endif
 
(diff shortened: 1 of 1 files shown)


cores/arduino/Arduino.h
```diff
@@ -53,7 +51,7 @@
 	     DT_PHA_BY_IDX(DT_PATH(zephyr_user), builtin_led_gpios, 0, pin)) > 0)
 #warning "pin not found in digital_pin_gpios"
 #else
-#define LED_BUILTIN                                                                                \
+#define LED_BUILTIN_                                                                                \
```
pillo79: I suggest using `ZARD_LED_BUILTIN` to make it clearer it's the Arduino core's definition.

pillo79 (approved):
LGTM! One small nit for your consideration.

cores/arduino/Arduino.h
soburi: Thank you for your point. Fixed.

pillo79 (approved):

cores/arduino/Arduino.h
```diff
@@ -67,15 +65,13 @@
 				    DT_PHA_BY_IDX(DT_ALIAS(led0), gpios, 0, pin)) > 0)
 #warning "pin not found in digital_pin_gpios"
 #else
-#define LED_BUILTIN                                                                                \
-	DIGITAL_PIN_GPIOS_FIND_PIN(DT_REG_ADDR(DT_PHANDLE_BY_IDX(DT_ALIAS(led0), gpios, 0)),     \
-				     DT_PHA_BY_IDX(DT_ALIAS(led0), gpios, 0, pin))
+#define ZARD_LED_BUILTIN                                                                           \
+	DIGITAL_PIN_GPIOS_FIND_PIN(DT_REG_ADDR(DT_PHANDLE_BY_IDX(DT_ALIAS(led0), gpios, 0)),           \
+							   DT_PHA_BY_IDX(DT_ALIAS(led0), gpios, 0, pin))
```
DhruvaG2000, line 70: I see lot's of whitespaces and tabs being mixed in github PR preview, can you stick with tabs please?

cores/arduino/Arduino.h
```diff
@@ -67,15 +65,13 @@
 				    DT_PHA_BY_IDX(DT_ALIAS(led0), gpios, 0, pin)) > 0)
 #warning "pin not found in digital_pin_gpios"
 #else
-#define LED_BUILTIN                                                                                \
-	DIGITAL_PIN_GPIOS_FIND_PIN(DT_REG_ADDR(DT_PHANDLE_BY_IDX(DT_ALIAS(led0), gpios, 0)),     \
-				     DT_PHA_BY_IDX(DT_ALIAS(led0), gpios, 0, pin))
+#define ZARD_LED_BUILTIN                                                                           \
```
DhruvaG2000, line 68: ZARD_ -> do you mean Zephyr Arduino or something? Can we think of slightly better naming here?

DhruvaG2000 (commented):
Functionality-wise we always had a `ifndef LED_BUILTIN` guarding this section so I am not expecting anything to change much. This approach does look cleaner now

cores/arduino/Arduino.h
soburi: This is the current .clang-format setting.
I'd like to consider this after reintegrating with the arduino-repo.

cores/arduino/Arduino.h
soburi: > ZARD_ -> do you mean Zephyr Arduino or something? Can we think of slightly better naming here?

Yes it is. I struggled with this, but couldn't come up with any other good ideas. If you have any suggestions, please let me know.
However, this isn't an externally visible interface; it simply defines a namespace specific to this module, so I don't think it's necessary to be too strict about it.
Since it often appears in long devicetree macros, it is important to consider that we do not want to use a very long prefix name.

pillo79: I think the idea was to keep with Zephyr using `Z_` for private stuff. Like Hiroshi, not a fan but can't come up with something better that's not way longer.

soburi: Z_ is Zephyr's namespace, so it's also difficult to use...

pillo79: Sorry I did not explain myself well. The comment was meant for Dhruva, I totally agree with using `ZARD_`, I wanted to say it _looks like_ the `Z_` that Zephyr uses for private stuff. I don't really like it, but have no better option.

soburi: Thank you for the explanation. It's clear that the conditions we cannot choice a very good name, but I think it's somewhat best in the choices. :-)

DhruvaG2000: ok I see... I am fine with this then

DhruvaG2000 (approved):

PR #153, opened 2026-02-01
cores: arduino: Bump to 0.53.1

soburi:
Bump `cores/arduino` to https://github.com/arduino/ArduinoCore-zephyr @0.53.1.

Including following features:

Including the following features:

- Support for Serial USB
- Support for DAC
- ITOA implementation
- C++ language support (abi/new)
- Multithread support
- UART Improvements

All of these are new features, and there is no downgrading of existing functionality.
It would be better to improve the #ifdef isolation in main.cpp, but there are no problems with operation as it is.


soburi:
@pillo79 @facchinm 

I have one request: as these files do not include an SPDX identifier, could we explicitly assign a license to them?
Zephyr is quite finicky about this sort of thing...

cores/arduino/time_macros.h
cores/arduino/overloads.h
cores/arduino/new.h
loader/bootanimation.h
loader/llext_exports.c
loader/fixups.c
loader/matrix.inc

pillo79:
Thank you SO MUCH for taking the time to do all of this. I always meant to start upstreaming but we've not at a stable point yet so it isn't easy to stop and look back. 

We started using `clang-format` as a code formatting tool and that complicated this effort even more. But I see this PR is using Zephyr style, please tell me you didn't have to change all the formatting by hand? :eyes:
(I always meant to raise this in Discord to agree on a format and just use that. Would you like to propose something?)

Thanks for pointing out the files with missing licenses - will fix them ASAP.

Is `cores/arduino` now perfectly synced with 0.53? I could start experimenting with using `ArduinoCore-zephyr` as a proper Zephyr module in our loader build, separating the two in a much cleaner way.

Again thanks for the effort! :clap:

soburi:
> We started using `clang-format` as a code formatting tool and that complicated this effort even more. But I see this PR is using Zephyr style, please tell me you didn't have to change all the formatting by hand? 👀 (I always meant to raise this in Discord to agree on a format and just use that. Would you like to propose something?)

I addressed only files that has diff here. I believe these gone into the Arduino style now.
IMO, after the merger, we'll also need to split up the all-in-one zephyrCommon.cpp.
So, this is a temporary issue, so I don't think it's something to worry about.
However, I'm concerned that the current .clang-format doesn't look very clean. We'll need to improve .clang-format with refer the zephyr's one and the rules a bit more.

> Thanks for pointing out the files with missing licenses - will fix them ASAP.


This is beyond my control, so I appreciate your help. Also, I think the issue under libraries needs to be discussed as a separate issue.


> Is `cores/arduino` now perfectly synced with 0.53? I could start experimenting with using `ArduinoCore-zephyr` as a proper Zephyr module in our loader build, separating the two in a much cleaner way.


With this patch series, there are almost no differences in core/arduino. However, there are a few patches that should be apply to arduino-repo from here. Since the loader build is independent, I don't think there's any problem with simply incorporating them, but I'll proceed with this in a separate PR.

PR #154, opened 2026-02-03
zephyr: module.yml: Update to support `click-through` property

soburi:
`click-through` has been added to the blobs properties,
so we will support this.
We will also update the `license-path` accordingly.

I also updated CI script.


soburi:
@DhruvaG2000 
I'm sorry, but could you please check this as well?
It seems that the behavior of `west blobs` has changed.

DhruvaG2000 (approved):

PR #160, opened 2026-02-06
libraries: loader: Bump to 0.53.1 (excluding blobs)

soburi:
Import the loader, excluding the blobs files.
The blobs files must be changed to retrieve them from west blobs.


soburi:
@pillo79 

```
loader/blobs/4343WA1_clm_blob.c
loader/blobs/wifi_nvram_image.h
loader/blobs/c33_bl.bin
```

It would be better to change these files to be retrieved using the `west blobs`.
However, I don't know the details of these files to write the config file, so if you know any, could you please let me know?

- Source-path (I've included the arduino-repo path for convenience, but if there is an "original" version, that would be preferable.)
- License (only for c33_bl.bin)
- Description (I've written what I can understand from the commit, but if you have more information, please let me know.)

PR #165, opened 2026-02-26
samples: hello_arduino: Correct printk format specifier (again)

soburi:
Fixed to use %zu which corresponds to size_t.


DhruvaG2000 (approved):
Ahh I read through https://docs.kernel.org/core-api/printk-formats.html and seems right, z it is!

PR #167, opened 2026-02-28
Backport few core fixes from Arduino fork

DhruvaG2000:
Added DNM while we first get clarity on what we should be doing about the missing SoB's .

### DIFF
--- a/cores/arduino/zephyrSerial.cpp
+++ b/cores/arduino/zephyrSerial.cpp
@@ -124,9 +124,9 @@ int arduino::ZephyrSerial::availableForWrite()
 {
 	int ret;
 
-	k_sem_take(&rx.sem, K_FOREVER);
-	ret = ring_buf_space_get(&rx.ringbuf);
-	k_sem_give(&rx.sem);
+	k_sem_take(&tx.sem, K_FOREVER);
+	ret = ring_buf_space_get(&tx.ringbuf);
+	k_sem_give(&tx.sem);
 
 	return ret;
 }
--- a/cores/arduino/zephyrSerial.cpp
+++ b/cores/arduino/zephyrSerial.cpp
@@ -60,6 +60,10 @@ void arduino::ZephyrSerial::begin(unsigned long baud, uint16_t conf) {
 
 	uart_configure(uart, &config);
 	uart_irq_callback_user_data_set(uart, arduino::ZephyrSerial::IrqDispatch, this);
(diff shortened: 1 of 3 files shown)


cores/arduino/zephyrSerial.cpp
```diff
@@ -116,6 +116,16 @@ int arduino::ZephyrSerial::available() {
 	return ret;
 }
 
+int arduino::ZephyrSerial::availableForWrite() {
+	int ret;
+
+	k_sem_take(&tx.sem, K_FOREVER);
+	ret = ring_buf_space_get(&tx.ringbuf);
+	k_sem_give(&tx.sem);
```
soburi, line 129: This part is from
https://github.com/arduino/ArduinoCore-zephyr/commit/66f8f71d3.

Shouldn't the contribution of the author of this commit also be listed?

DhruvaG2000: Not sure I understand , the author is indeed kept as is, it's just a cherrypick.

Image

soburi: https://github.com/arduino/ArduinoCore-zephyr/blame/main/cores/arduino/zephyrSerial.cpp#L126-L134

As you can clearly understand from blame list, this function
has three authorships.

https://github.com/arduino/ArduinoCore-zephyr/commit/66f8f71d398db7b19da5d492491412f0855c89ba

This also only has matched three lines.

DhruvaG2000: Ah okay, so I will try to pick all 3 then

DhruvaG2000: One of them was just a formatting fix, which I'm leaving out for now

cores/arduino/zephyrSerial.cpp
```diff
@@ -116,24 +120,36 @@ int arduino::ZephyrSerial::available() {
...
+{
 	uint8_t data;
 
 	k_sem_take(&rx.sem, K_FOREVER);
-	ring_buf_peek(&rx.ringbuf, &data, 1);
+	uint32_t cb_ret = ring_buf_peek(&rx.ringbuf, &data, 1);
 	k_sem_give(&rx.sem);
 
-	return data;
+	return cb_ret? data : -1;
```
soburi, line 142: The content appears fine, but there are a few formatting issues.

PR #172, opened 2026-04-03
variants: Add initial native_sim support

DhruvaG2000:
Add variants/native_sim/ with native_sim.overlay and native_sim_pinmap.h to support the native_sim board (renamed from native_posix in Zephyr).

native_sim support enables:
- CI/testing without requiring physical hardware
- Rapid development iteration on host machine
- Debugging Arduino API implementations with native toolchains
- Verifying devicetree overlays and pin mappings in simulation
- Running samples like hello_arduino locally before flashing targets


soburi (changes requested):
Please add SPDX-Identifier.

soburi (approved):

PR #174, opened 2026-04-17
Fix micros() for nano without breaking giga

soburi:
The rollover fix for the giga used the 64 bit timer instead of 32 bits.  The nano does not have a 64 bit timer, so you need to use the 32 bit instead.

So I know check to see if the processor supports 64 bits if so use it else fall back to 32 bit.

Note: This branch doesn't include support for Arduino Giga,
      but we'll include it as it's a reasonable solution for
      preventing timer overflows.

### DIFF
--- a/cores/arduino/zephyrCommon.cpp
+++ b/cores/arduino/zephyrCommon.cpp
@@ -347,7 +347,11 @@ void delayMicroseconds(unsigned int us) {
 }
 
 unsigned long micros(void) {
+#ifdef CONFIG_TIMER_HAS_64BIT_CYCLE_COUNTER
+	return k_cyc_to_us_floor32(k_cycle_get_64());
+#else
 	return k_cyc_to_us_floor32(k_cycle_get_32());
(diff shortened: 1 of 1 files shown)


cores/arduino/zephyrCommon.cpp
```diff
@@ -347,7 +347,11 @@ void delayMicroseconds(unsigned int us) {
 }
 
 unsigned long micros(void) {
+#ifdef CONFIG_TIMER_HAS_64BIT_CYCLE_COUNTER
```
DhruvaG2000, line 350: Where does this config get enabled? I don't see it anywhere unless I am missing something.

soburi: https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/timer/Kconfig#L12-L16

This is pegged to the Zephyr configuration.

DhruvaG2000 (approved):

PR #188, opened 2026-05-06
cores: arduino: Split zephyrCommon.cpp

soburi:
Split zephyrCommon.cpp into small cpp files.


DhruvaG2000 (approved):
A much needed change to split this file up! 
I can see it's already merged in the fork: https://github.com/arduino/ArduinoCore-zephyr/commits/main/cores/arduino/zephyrCommon.cpp so on a high level view it looks good to me as well.

PR #201, opened 2026-07-04
variants: import Nano 33 BLE Sense updates from 0.55.0

soburi:
Bring in the Nano 33 BLE Sense variant changes from the 0.55.0 delta, including board configuration, extended pin mappings, peripheral setup, sensor power initialization, and example skip metadata.

USB has not yet been implemented in this repository, cdc_acm_uart is disabled.

This is a temporary measure; additional modifications are required to allow the Nano 33 BLE and Nano 33 BLE Sense to coexist.


variants/arduino_nano_33_ble_nrf52840_sense/arduino_nano_33_ble_nrf52840_sense.overlay
```diff
@@ -1,5 +1,69 @@
...
+
+/{
+//	chosen {
+//		zephyr,log-uart = &log_uarts;
+//	};
+//
+//	log_uarts: log_uarts {
+//		compatible = "zephyr,log-uart";
+//		uarts = <&uart0 &board_cdc_acm_uart>;
+//	};
```
DhruvaG2000: clean this up please?

DhruvaG2000 (changes requested):
I see some open comments from copilot as well, please take a look and resolve

variants/arduino_nano_33_ble_nrf52840_sense/arduino_nano_33_ble_nrf52840_sense.overlay
soburi: Since USB enablement is also part of the series of PRs, I will handle this after completing those. That would come expected results.

PR #203, opened 2026-07-05
core: Fixes for proper core compatibility

soburi:
Adding include math.h for improve core compatibility.


DhruvaG2000 (approved):
Just fix the commit title its a trivial change and it should describe just that.

The description can have the part where its fixing and what it is fixing.

I hope you will address that and then merge, thnx

PR #213, opened 2026-07-05
variants: Restore removed empty variant.h

soburi:
It was removed it in 2044d497, but I am reverting the change because it actually makes management more complicated.

### DIFF
--- a/variants/arduino_mkrzero/variant.h
+++ b/variants/arduino_mkrzero/variant.h
@@ -1,0 +1,5 @@
+/*
+ * Copyright (c) 2022 Dhruva Gole
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
--- a/variants/nrf52840dk_nrf52840/variant.h
+++ b/variants/nrf52840dk_nrf52840/variant.h
@@ -1,0 +1,5 @@
+/*
+ * Copyright (c) 2022 Mike Szczys
+ *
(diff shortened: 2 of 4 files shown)


DhruvaG2000 (commented):
I'm sorry could you elaborate a bit more?
> it actually makes management more complicated.

More complex in what case/ scenario if you could point me to it?

soburi:
> More complex in what case/ scenario if you could point me to it?

Since this is included due to the Arduino IDE, it creates a discrepancy with the Arduino-side implementation.
While its presence or absence does not affect functionality for a Zephyr native build, I believe it is preferable to avoid such differences for the sake of maintainability, especially if the same code is to be used with the Arduino IDE.
Currently, features like rpi_pico, which are not yet available in the Arduino IDE, will likely have their settings incorporated for the Arduino IDE in the near future.

DhruvaG2000 (approved):
approving, but as a co-maintainer do you think we should mark this as temporary or HACK commit that we need to come back to later on at some point?

soburi:
> approving, but as a co-maintainer do you think we should mark this as temporary or HACK commit that we need to come back to later on at some point?

Since this itself doesn't cause any adverse effects, I don't think there's a need to mention it. I believe we'll need to add support for Arduino once things take shape to some extent, so the issue should be resolved at that point.

